PR #26, opened 2016-07-26
Ported over Convert Reads to File; added exporter function for download

msneddon:
Code mostly copied from https://github.com/MrCreosote/kb_read_library_to_file


MrCreosote:
Since the uploader supports bzipped files, the downloader needs to as well.

MrCreosote:
The download functions should be rewritten to use DataFileUtils rather than duplicating their functionality.

MrCreosote:
There needs to be a way for the user to know which reads are the fwd / rev reads

EDIT - never mind - the function appends the file names with .fwd.fastq or .rev.fastq

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -156,6 +183,463 @@ def validateFASTA(self, ctx, params):
...
+        self.log('Interleaving files {} and {} to {}'.format(
+            fwdpath, revpath, targetpath))
+        with open(targetpath, 'w') as t:
+            with open(fwdpath, 'r') as f, open(revpath, 'r') as r:
+                while True:
+                    line = f.readline()
+                    # since FASTQ cannot contain blank lines
+                    if not line or not line.strip():
+                        break
+                    t.write(line.strip() + '\n')
```
MrCreosote, line 359: Currently the uploader allows blank lines in fq. Either this needs to handle & ignore blank lines or the uploader needs to fix them and reupload the files.

Same for the deinterleave fn above.

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -156,6 +183,463 @@ def validateFASTA(self, ctx, params):
...
+                shockfile, gzip, isgz, self.get_file_prefix() + '.inter.fastq')
+        else:
+            if isgz:
+                # we expect the job runner to clean up for us
+                shockfile = self.gunzip(shockfile)
+            fwdpath = os.path.join(self.scratch, self.get_file_prefix() +
+                                   '.fwd.fastq')
+            revpath = os.path.join(self.scratch, self.get_file_prefix() +
+                                   '.rev.fastq')
+            self.deinterleave(shockfile, fwdpath, revpath)
```
MrCreosote, line 446: These look like bugs - get_file_prefix already prepends self.scratch

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -156,6 +183,463 @@ def validateFASTA(self, ctx, params):
...
+
+        ret = {}
+        if interleave:
+            # we expect the job runner to clean up for us
+            if fwdisgz:
+                fwdshock = self.gunzip(fwdshock)
+            if revisgz:
+                revshock = self.gunzip(revshock)
+            intpath = os.path.join(self.scratch, self.get_file_prefix() +
+                                   '.inter.fastq')
```
MrCreosote, line 474: Same here - no need to prepend self.scratch

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -411,6 +906,250 @@ def upload_reads(self, ctx, params):
...
+        read_libraries = [ params['input_ref'] ]
+        read_lib_info = self.convert_read_library_to_file(ctx, { 
+                              'read_libraries': read_libraries
+                            })[0]['files']
+        pprint(read_lib_info)
+
+        # create the output directory and move the file there
+        export_package_dir = os.path.join(self.scratch, info[1])
+        os.makedirs(export_package_dir)
+        for r in read_lib_info:
```
MrCreosote, line 1111: this loop seems to be unnecessary since there's only one entry in read_lib_info and r = params['input_ref']

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -411,6 +906,250 @@ def upload_reads(self, ctx, params):
...
+            if 'inter' in lib['files'] and lib['files']['inter'] is not None and lib['files']['inter'] is not '':
+                print('packaging interleaved reads file: '+f_name)
+                f_name = lib['files']['inter']
+                shutil.move(f_name, os.path.join(export_package_dir, os.path.basename(f_name)))
+
+            if 'sing' in lib['files'] and lib['files']['sing'] is not None and lib['files']['sing'] is not '':
+                print('packaging single end reads file: '+f_name)
+                f_name = lib['files']['sing']
+                shutil.move(f_name, os.path.join(export_package_dir, os.path.basename(f_name)))   
+
```
MrCreosote, line 1133: If the filetype is in the return structure, it can't be None or empty.

MrCreosote:
export tests are pretty incomplete - just one happy path.

PR #27, opened 2016-09-29
Reads downloader

MrCreosote:
Ported, with significant alterations, from https://github.com/MrCreosote/kb_read_library_to_file

Uses DFU to interact with the WS and Shock


msneddon:
Interface looks good to me.  Note that we'll need to add one more method to wrap the downloader for connecting this to the narrative download.

MrCreosote:
Merged on Mike's behalf since he apparently can't merge here.

PR #31, opened 2016-10-20
Make all uploaded paired end reads  to be interleaved

jkbaumohl:
Changes to uploader and tests to make all paired non interleaved on input stored as interleaved in the PairedEnd reads object


MrCreosote:
Against master, going to move to develop

PR #32, opened 2016-10-20
Paired End upload now interleaves if 2 files and does EAUtils stats

jkbaumohl:
Changes to uploader and tests to make all paired non interleaved on input stored as interleaved in the PairedEnd reads object


lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -388,12 +388,16 @@ def _read_fq_record(self, source_obj_ref, source_obj_name,
...
+                        error_message_bindings.insert(0,source_obj_ref)
+                        error_message_bindings.insert(0,source_obj_name)
+                    error_message += 'Shock node {}, Shock filename {}'
                     raise ValueError(
-                        ('Reading FASTQ record failed - non-blank lines ' +
-                         'are not a multiple of four. Workspace reads ' +
-                         'object {} ({}), Shock node {}, Shock filename {}')
-                        .format(source_obj_name, source_obj_ref,
-                                shock_node, shock_filename))
+                        (error_message)
```
MrCreosote: You don't need the params around error_message here - they were there originally so the .format() call applied to the whole string rather than just the last line.

MrCreosote:
General comment: there are multiple pep8 violations across the changes, too many to note each individually. Run a pep8 checker and fix the violations.

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -416,15 +420,17 @@ def interleave(self, source_obj_ref, source_obj_name, fwd_shock_filename,
...
-                            ('Interleave failed - reads files do not have ' +
-                             'an equal number of records. Workspace reads ' +
-                             'object {} ({}), ' +
-                             'forward Shock node {}, filename {}, ' +
-                             'reverse Shock node {}, filename {}')
-                            .format(source_obj_name, source_obj_ref,
-                                    fwd_shock_node, fwd_shock_filename,
-                                    rev_shock_node, rev_shock_filename))
+                            (error_message
+                            .format(*error_message_bindings)))
```
MrCreosote: Extraneous parens

MrCreosote:
Tests pass locally

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,6 +829,44 @@ def upload_reads(self, ctx, params):
         self.log('Starting upload reads, parsing args')
         o, wsid, name, objid, kbtype, single_end, fwdid, revid, shock = (
             self._proc_upload_reads_params(params))
+
+        needs_to_be_interleaved = 0
+        if not single_end and revid is not None:
+            needs_to_be_interleaved = 1
```
MrCreosote: needs_to_be_interleaved is never used

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,6 +829,44 @@ def upload_reads(self, ctx, params):
...
+
+        if revid and shock:
+            #Download files so we can interleave them later.
+            fileinput = [{'shock_id': fwdid,
+                          'file_path': self.scratch + '/fwd/',
+                          'unpack': 'uncompress'}]
+            fileinput.append({'shock_id': revid,
+                              'file_path': self.scratch + '/rev/',
+                              'unpack': 'uncompress'})
+            dfu = DataFileUtil(self.callback_url)
```
MrCreosote: move the dfu declaration outside the if block and remove the dfu declaration on line 903, just declare it once

test/ReadsUtils_server_test.py
```diff
@@ -751,19 +751,19 @@ def test_paired_end_reads(self):
...
         self.assertEqual(d['insert_size_mean'], None)
         self.assertEqual(d['insert_size_std_dev'], None)
-        self.check_lib(d['lib1'], 1116, 'Sample5_noninterleaved.1.fastq',
-                       ret1['id'], '140a61c7f183dd6a2b93ef195bb3ec63')
-        self.check_lib(d['lib2'], 2847, 'Sample1.fastq.gz',
-                       ret2['id'], '48efea6945c4382c68f5eac485c177c2')
+#        self.check_lib(d['lib1'], 3051931, 'small.forward.fq',
+#                       ret1['id'], 'e7dcea3e40d73ca0f71d11b044f30ded')
+#        self.check_lib(d['lib2'], 3051932, 'small.reverse.fq',
+#                       ret2['id'], '2cf41e49cd6b9fdcf1e511b083bb42b5')
```
MrCreosote, line 810: The check_lib method still needs to be run, the inputs just need to be updated to match the new file

test/ReadsUtils_server_test.py
```diff
@@ -791,14 +791,14 @@ def test_paired_end_reads_file(self):
...
         self.assertEqual(d['insert_size_mean'], None)
         self.assertEqual(d['insert_size_std_dev'], None)
-        self.check_lib(d['lib1'], 604, 'Sample5_noninterleaved.1.fastq.gz',
-                       node1, None)
-        self.check_lib(d['lib2'], 2835, 'Sample1.fastq.gz',
-                       node2, None)
+#        self.check_lib(d['lib1'], 6103863, 'small.forward.fq',
+#                       node1, None)
+#        self.check_lib(d['lib2'], 3051932, 'small.reverse.fq',
+#                       node2, None)
```
MrCreosote: Same, the lib still needs to be checked

test/ReadsUtils_server_test.py
```diff
@@ -1054,21 +1054,6 @@ def test_upload_fail_interleaved_for_single(self):
...
-             'fwd_id': ret1['id'],
-             'rev_id': ret2['id'],
-             'name': 'bar'
-             },
-            'Invalid fasta file /kb/module/work/tmp/rev/Sample5_interleaved' +
-            '.fastq from Shock node ' + ret2['id'])
-        self.delete_shock_node(ret1['id'])
-        self.delete_shock_node(ret2['id'])
-
     # Download tests ########################################################
```
MrCreosote, line 1277: There should be a tests that provide uninterleavable input in shock nodes and local files that covers both bad individual records and mismatched files

MrCreosote (dismissed):

MrCreosote:
One other small point is that the documentation needs to be updated to note the new uploader behavior

MrCreosote (dismissed):
Tests fail immediately:

```
bash test/run_tests.sh
E
======================================================================
ERROR: ReadsUtils_server_test (unittest.loader.ModuleImportFailure)
----------------------------------------------------------------------
ImportError: Failed to import test module: ReadsUtils_server_test
Traceback (most recent call last):
  File "/usr/lib/python2.7/unittest/loader.py", line 254, in _find_tests
    module = self._get_module_from_name(name)
  File "/usr/lib/python2.7/unittest/loader.py", line 232, in _get_module_from_name
    __import__(name)
  File "ReadsUtils_server_test.py", line 24, in <module>
    from ReadsUtils.ReadsUtilsImpl import ReadsUtils
  File "/kb/module/lib/ReadsUtils/ReadsUtilsImpl.py", line 11, in <module>
    from kb_ea_utils.kb_ea_utilsClient import kb_ea_utils
ImportError: No module named kb_ea_utils.kb_ea_utilsClient


----------------------------------------------------------------------
Ran 1 test in 0.000s

FAILED (errors=1)
```

Also there are many pep8 violations and the documentation still needs to be updated to account for the new behavior.

 I'll continue the review when this stuff is fixed because these changes may be extensive.

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -625,6 +598,56 @@ def validateFASTA(self, ctx, params):
...
+                                     'len stdev':'read_length_stdev',
+                                     'phred':'phred_type',
+                                     'dups':'number_of_duplicates',
+                                     'qual min':'qual_min',
+                                     'qual max':'qual_max',
+                                     'qual mean':'qual_mean',
+                                     'qual stdev':'qual_stdev'}
+        integer_fields = ['read_count','read_size','number_of_duplicates']
+        for line in report_lines:
+            line_elements = line.split()
```
MrCreosote, line 614: This looks like the eautils wrapper just returns a string with the contents of what is presumably stdout or file output from eautils. eautils should be modified to do the parsing itself and return a sane data structure with that information. There's no reason for the eautils wrapper to force everyone that uses it to do the parsing themselves.

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -125,7 +126,7 @@ def _proc_upload_reads_params(self, params):
             kbtype = 'KBaseFile.PairedEndLibrary'
             single_end = False
         if revid:
-            interleaved = 0
+            interleaved = 1
```
MrCreosote: This can just be merged into the original interleaved line, something like
interleaved = 1 if params.get('interleaved') or revid else 0

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,15 +846,54 @@ def upload_reads(self, ctx, params):
...
         else:
-            fhandle, rhandle, fsize, rsize = self.process_files_for_upload(
-                fwdid, revid, interleaved)
+            #Not shock
+            fwdpath = fwdid
+            revpath = revid
+            fwdname = None
+            revname = None
+
+        shock = False
```
MrCreosote: No need for this line, shock is never used again

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,15 +846,54 @@ def upload_reads(self, ctx, params):
         self.log('Starting upload reads, parsing args')
         o, wsid, name, objid, kbtype, single_end, fwdid, revid, shock = (
             self._proc_upload_reads_params(params))
-        interleaved = 1 if (not single_end and not revid) else 0
+        #IF from shock fwdid and revid are shock nodes, if not shock they are file paths
+        dfu = DataFileUtil(self.callback_url)
+        fwdpath = None
+        revpath = None
```
MrCreosote, line 846: No need for these two lines, fwd and revpath are always set in the if/else below

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,15 +846,54 @@ def upload_reads(self, ctx, params):
...
+                                    '.inter.fastq')
+            self.interleave(
+                    None, None, fwdname, fwdpath,
+                    revname, revpath, fwdpath, revpath, intpath)
+            fwdpath = intpath
+            revid = None
+            revpath = None
+  
+        interleaved = 1 if (not single_end and not revid) else 0        
+
```
MrCreosote, line 871: revid is guaranteed to be false here, no need to check

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,15 +846,54 @@ def upload_reads(self, ctx, params):
...
+        shock = False
+        if revid:
+            #now interleave the files
+            intpath = os.path.join(self.scratch, self.get_file_prefix() +
+                                    '.inter.fastq')
+            self.interleave(
+                    None, None, fwdname, fwdpath,
+                    revname, revpath, fwdpath, revpath, intpath)
+            fwdpath = intpath
+            revid = None
```
MrCreosote, line 870: No need to set revid to none, see below. Just always set interleaved to 1 if it's not single end

MrCreosote (changes requested):
Non-line specific comments:
- We need to resolve what reads_size really means and update the type. I sent an email about this to Mike & Jason yesterday.
- Still many pep8 violations
- The documentation still needs to be updated to explain the new behavior
- The downloader needs to be updated to return the new fields in the object
- ~~Tests still fail.~~ Nope, they failed because the tests default to appdev and I forgot to change to ci. Will review test files in a new review.

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -625,6 +598,56 @@ def validateFASTA(self, ctx, params):
         # return the results
         return [out]
 
+    def calculate_fq_stats(self, reads_object, file_path):
+        eautils = kb_ea_utils(self.callback_url)
```
MrCreosote: The ea utils client is installed so it always points to dev. This is bad practice - it should be parameterized so it can be easily changed when this module is out of development mode (or put in the deploy file).

lib/kb_ea_utils/kb_ea_utilsClient.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+class kb_ea_utils(object):
+
+    def __init__(
+            self, url=None, timeout=30 * 60, user_id=None,
+            password=None, token=None, ignore_authrc=False,
+            trust_all_ssl_certificates=False,
+            auth_svc='https://kbase.us/services/authorization/Sessions/Login',
+            service_ver='dev',
```
MrCreosote: Reinstall client without specifying a service ver

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,15 +846,54 @@ def upload_reads(self, ctx, params):
...
+                    None, None, fwdname, fwdpath,
+                    revname, revpath, fwdpath, revpath, intpath)
+            fwdpath = intpath
+            revid = None
+            revpath = None
+  
+        interleaved = 1 if (not single_end and not revid) else 0        
+
+        fhandle, rhandle, fsize, rsize = self.process_files_for_upload(
+                fwdpath, revpath, interleaved)
```
MrCreosote: OK, there's actually a bigger problem here - the code now always uses this function to upload file to shock,which assumes that the files were local. That means any error messages from the function don't include the shock IDs from which the files originate, which makes it impossible to debug errors.

test/ReadsUtils_server_test.py
```diff
@@ -641,17 +640,17 @@ def test_forward_reads_file(self):
                        'name': 'filereads1'})
         obj = self.dfu.get_objects(
             {'object_refs': [self.ws_info[1] + '/filereads1']})['data'][0]
-        node = obj['data']['lib']['file']['id']
-        self.delete_shock_node(node)
```
MrCreosote, line 645: The shock node still needs to be deleted when the test is done with it

test/ReadsUtils_server_test.py
```diff
@@ -641,17 +640,17 @@ def test_forward_reads_file(self):
...
+    
         self.assertEqual(ref[0]['obj_ref'], self.make_ref(obj['info']))
         self.assertEqual(obj['info'][2].startswith(
                         'KBaseFile.SingleEndLibrary'), True)
         d = obj['data']
+        node = d["lib"]["file"]["id"]
```
MrCreosote: node is never used

test/ReadsUtils_server_test.py
```diff
@@ -747,23 +759,22 @@ def test_paired_end_reads(self):
         self.assertEqual(obj['info'][2].startswith(
                         'KBaseFile.PairedEndLibrary'), True)
         d = obj['data']
+        file_name = d["lib1"]["file"]["file_name"]
```
MrCreosote, line 797: Should check the filename is as you expect (at least the suffix)

test/ReadsUtils_server_test.py
```diff
@@ -775,30 +786,44 @@ def test_paired_end_reads_file(self):
...
                        'name': 'pairedreadsfile1',
-                       'interleaved': 1})
+                       'interleaved': 0})
         obj = self.dfu.get_objects(
             {'object_refs': [self.ws_info[1] + '/pairedreadsfile1']}
         )['data'][0]
-        node1 = obj['data']['lib1']['file']['id']
-        node2 = obj['data']['lib2']['file']['id']
-        self.delete_shock_node(node1)
-        self.delete_shock_node(node2)
```
MrCreosote, line 785: Shock nodes still need to be deleted after the test is done with them

test/ReadsUtils_server_test.py
```diff
@@ -775,30 +786,44 @@ def test_paired_end_reads_file(self):
...
         self.assertEqual(ref[0]['obj_ref'], self.make_ref(obj['info']))
         self.assertEqual(obj['info'][2].startswith(
                         'KBaseFile.PairedEndLibrary'), True)
         d = obj['data']
+        node1 = d["lib1"]["file"]["id"]
+        file_name = d["lib1"]["file"]["file_name"]
```
MrCreosote, line 834: Should check the file name is as you expect

test/ReadsUtils_server_test.py
```diff
@@ -1027,8 +1079,9 @@ def test_upload_fail_bad_fastq(self):
              'fwd_id': ret['id'],
              'name': 'bar'
              },
-            'Invalid fasta file /kb/module/work/tmp/fwd/Sample1_invalid' +
-            '.fastq from Shock node ' + ret['id'])
+            'Invalid fasta file /kb/module/work/tmp/fwd/Sample1_invalid.fastq')
+#            'Invalid fasta file /kb/module/work/tmp/fwd/Sample1_invalid' +
+#            '.fastq from Shock node ' + ret['id'])
```
MrCreosote: Not having the source shock node here makes errors impossible to debug

test/ReadsUtils_server_test.py
```diff
@@ -1050,25 +1103,11 @@ def test_upload_fail_interleaved_for_single(self):
              'fwd_id': ret['id'],
              'name': 'bar'
              },
-            'Invalid fasta file /kb/module/work/tmp/fwd/Sample5_interleaved' +
-            '.fastq from Shock node ' + ret['id'])
+            'Invalid fasta file /kb/module/work/tmp/fwd/Sample5_interleaved.fastq')
+#            'Invalid fasta file /kb/module/work/tmp/fwd/Sample5_interleaved' +
+#            '.fastq from Shock node ' + ret['id'])
```
MrCreosote: Not having the source shock node here makes errors impossible to debug

test/ReadsUtils_server_test.py
```diff
@@ -1050,25 +1103,11 @@ def test_upload_fail_interleaved_for_single(self):
...
-             'fwd_id': ret1['id'],
-             'rev_id': ret2['id'],
-             'name': 'bar'
-             },
-            'Invalid fasta file /kb/module/work/tmp/rev/Sample5_interleaved' +
-            '.fastq from Shock node ' + ret2['id'])
-        self.delete_shock_node(ret1['id'])
-        self.delete_shock_node(ret2['id'])
-
     # Download tests ########################################################
```
MrCreosote, line 1277: There's still no tests covering the case of providing f and r reads in shock nodes that can't be interleaved; all the possible failure modes for interleaving need to be covered

MrCreosote (dismissed):
Review #2. There's now two active reviews, just to be clear - this one and the previous one.

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: Same comment as previous review: 
This can just be merged into the original interleaved line, something like
interleaved = 1 if params.get('interleaved') or revid else 0

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,15 +850,51 @@ def upload_reads(self, ctx, params):
         self.log('Starting upload reads, parsing args')
         o, wsid, name, objid, kbtype, single_end, fwdid, revid, shock = (
             self._proc_upload_reads_params(params))
-        interleaved = 1 if (not single_end and not revid) else 0
+        #IF from shock fwdid and revid are shock nodes, if not shock they are file paths
+        dfu = DataFileUtil(self.callback_url)
+        fwdpath = None
+        revpath = None
```
MrCreosote, line 846: Same comment as previous review:
No need for these two lines, fwd and revpath are always set in the if/else below

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,15 +850,51 @@ def upload_reads(self, ctx, params):
...
+            fileinput = [{'shock_id': fwdid,
+                          'file_path': self.scratch + '/fwd/',
+                          'unpack': 'uncompress'}]
+            if revid:
+                fileinput.append({'shock_id': revid,
+                                  'file_path': self.scratch + '/rev/',
+                                  'unpack': 'uncompress'})
+            self.log('downloading reads file(s) from Shock')
+            files = dfu.shock_to_file_mass(fileinput)
+            shock = False
```
MrCreosote: No need for this line, shock is never used again

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,15 +850,51 @@ def upload_reads(self, ctx, params):
...
+                revname = files[1]["node_file_name"]
         else:
-            fhandle, rhandle, fsize, rsize = self.process_files_for_upload(
-                fwdid, revid, interleaved)
+            #Not shock
+            fwdpath = fwdid
+            revpath = revid
+            fwdname = None
+            revname = None
+        shock = False
```
MrCreosote: No need for this line, shock is never used again

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,15 +850,51 @@ def upload_reads(self, ctx, params):
...
+        if revid:
+            #now interleave the files
+            intpath = os.path.join(self.scratch, self.get_file_prefix() + '.inter.fastq')
+            self.interleave(None, None, fwdname, fwdpath,
+                            revname, revpath, fwdpath, revpath, intpath)
+            fwdpath = intpath
+            revid = None
```
MrCreosote, line 870: No need to set revid to none, see below. Just always set interleaved to 1 if it's not single end

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,15 +850,51 @@ def upload_reads(self, ctx, params):
...
+            revpath = None
+
+        interleaved = 1 if (not single_end and not revid) else 0
```
MrCreosote: revid is guaranteed to be false here, no need to check

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -625,6 +602,56 @@ def validateFASTA(self, ctx, params):
         # return the results
         return [out]
 
+    def calculate_fq_stats(self, reads_object, file_path):
+        eautils = kb_ea_utils(self.callback_url, service_ver='dev')
```
MrCreosote, line 597: Add a #TODO to remind you to remove sevice_ver='dev' before releasing this version

test/ReadsUtils_server_test.py
```diff
@@ -747,23 +761,22 @@ def test_paired_end_reads(self):
         self.assertEqual(obj['info'][2].startswith(
                         'KBaseFile.PairedEndLibrary'), True)
         d = obj['data']
+        file_name = d["lib1"]["file"]["file_name"]
+        self.assertTrue(file_name.endswith("fastq.gz"))
```
MrCreosote: Should check for .inter.fastq.qz, not just .fastq.gz, IIUC

test/ReadsUtils_server_test.py
```diff
@@ -775,30 +788,41 @@ def test_paired_end_reads_file(self):
...
                        'name': 'pairedreadsfile1',
-                       'interleaved': 1})
+                       'interleaved': 0})
         obj = self.dfu.get_objects(
             {'object_refs': [self.ws_info[1] + '/pairedreadsfile1']}
         )['data'][0]
-        node1 = obj['data']['lib1']['file']['id']
-        node2 = obj['data']['lib2']['file']['id']
-        self.delete_shock_node(node1)
-        self.delete_shock_node(node2)
```
MrCreosote, line 785: Same comment as previous review:
Shock nodes still need to be deleted after the test is done with them

test/ReadsUtils_server_test.py
```diff
@@ -775,30 +788,41 @@ def test_paired_end_reads_file(self):
...
         self.assertEqual(ref[0]['obj_ref'], self.make_ref(obj['info']))
         self.assertEqual(obj['info'][2].startswith(
                         'KBaseFile.PairedEndLibrary'), True)
         d = obj['data']
+        file_name = d["lib1"]["file"]["file_name"]
```
MrCreosote, line 834: Should check file name is as you expect

test/ReadsUtils_server_test.py
```diff
@@ -1027,8 +1078,7 @@ def test_upload_fail_bad_fastq(self):
              'fwd_id': ret['id'],
              'name': 'bar'
              },
-            'Invalid fasta file /kb/module/work/tmp/fwd/Sample1_invalid' +
-            '.fastq from Shock node ' + ret['id'])
+            'Invalid fasta file /kb/module/work/tmp/fwd/Sample1_invalid.fastq')
```
MrCreosote: Not having the source shock node here makes errors impossible to debug

test/ReadsUtils_server_test.py
```diff
@@ -1050,22 +1100,41 @@ def test_upload_fail_interleaved_for_single(self):
              'fwd_id': ret['id'],
              'name': 'bar'
              },
-            'Invalid fasta file /kb/module/work/tmp/fwd/Sample5_interleaved' +
-            '.fastq from Shock node ' + ret['id'])
+            'Invalid fasta file /kb/module/work/tmp/fwd/Sample5_interleaved.fastq')
```
MrCreosote: Not having the source shock node here makes errors impossible to debug

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,15 +850,51 @@ def upload_reads(self, ctx, params):
...
+            self.interleave(None, None, fwdname, fwdpath,
+                            revname, revpath, fwdpath, revpath, intpath)
+            fwdpath = intpath
+            revid = None
+            revpath = None
+
+        interleaved = 1 if (not single_end and not revid) else 0
+
+        fhandle, rhandle, fsize, rsize = \
+            self.process_files_for_upload(fwdpath, revpath, interleaved)
```
MrCreosote: OK, there's actually a bigger problem here - the code now always uses this function to upload file to shock,which assumes that the files were local. That means any error messages from the function don't include the shock IDs from which the files originate, which makes it impossible to debug errors.

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -325,9 +295,11 @@ def _download_reads_from_shock(self, ref, obj_name, handle, file_type):
...
                   'file_path': os.path.join(self.scratch, handle['id'])
                   }
-        # TODO LATER may want to do dl en masse, but that means if there's a bad file it won't be caught until everythings dl'd @IgnorePep8
-        # TODO LATER add method to DFU to get shock attribs and check filename prior to download @IgnorePep8
-        # TODO LATER at least check handle filename & file type are ok before download @IgnorePep8
+        # TODO LATER may want to do dl en masse, but that means if there's
+        # TODO a bad file it won't be caught until everythings dl'd @IgnorePep8
+        # TODO LATER add method to DFU to get shock attribs and check filename prior to
+        # TODO download LATER at least check handle filename & file type are
+        # TODO ok before download @IgnorePep8
```
MrCreosote: TODOs need to be all one line or parsers will think there are 5 TODOs rather than 3. the @IgnorePep8 suffixes are there for this reason

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -625,6 +602,56 @@ def validateFASTA(self, ctx, params):
         # return the results
         return [out]
 
+    def calculate_fq_stats(self, reads_object, file_path):
+        eautils = kb_ea_utils(self.callback_url, service_ver='dev')
+        ea_report = eautils.get_ea_utils_stats({'read_library_path': file_path})
+#        print "Full Report : {}".format(ea_report)
+
+        report_lines = ea_report.splitlines()
```
MrCreosote, line 601: EA utils should not force the user to parse the output of the ea utils program - that should be the the job of the SDK method. At minimum this parsing code should be moved inside the SDK method. Otherwise we're going to reproduce this parsing code over and over.

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -345,7 +317,8 @@ def _download_reads_from_shock(self, ref, obj_name, handle, file_type):
                          'object {} ({}). Shock node {}')
                         .format(n, f, obj_name, ref, handle['id']))
                 ok = True
-        # TODO this is untested. You have to try pretty hard to upload a file without a name to Shock. @IgnorePep8
+        # TODO this is untested. You have to try pretty hard to upload a file
+        # TODO without a name to Shock.
```
MrCreosote: Same, this is now 2 TODOs rather than one

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -743,7 +769,8 @@ def upload_reads(self, ctx, params):
            parameters: rev_id - the shock node id containing the
            reverse/right reads for paired end, non-interleaved reads. - OR -
            rev_file - a local path to the reads data file containing the
-           reverse/right reads for paired end, non-interleaved reads.
+           reverse/right reads for paired end, non-interleaved reads, note the
+           reverse file will get interleaved with the forward file.
```
MrCreosote: This documentation is generated from the spec file, and so will get overwritten the next time the code is recompiled. The changes need to be made in the spec file.

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,15 +850,51 @@ def upload_reads(self, ctx, params):
...
+            fwdpath = fwdid
+            revpath = revid
+            fwdname = None
+            revname = None
+        shock = False
+        if revid:
+            #now interleave the files
+            intpath = os.path.join(self.scratch, self.get_file_prefix() + '.inter.fastq')
+            self.interleave(None, None, fwdname, fwdpath,
+                            revname, revpath, fwdpath, revpath, intpath)
```
MrCreosote: This has a similar problem to process_files_for_upload - if there's an error, the error won't contain the shock IDs, which makes debugging really hard.

test/ReadsUtils_server_test.py
```diff
@@ -1050,22 +1100,41 @@ def test_upload_fail_interleaved_for_single(self):
...
+                                'sequencing_tech': 'seqtech-pr1',
+                                'wsname': self.ws_info[1],
+                                'name': 'pairedreads1',
+                                'interleaved': 0},
+                               'Interleave failed - reads files do not have ' +
+                               'an equal number of records. forward Shock node ' +
+                               '/kb/module/work/tmp/fwd/small.forward.fq, ' +
+                               'filename small.forward.fq, reverse Shock node ' +
+                               '/kb/module/work/tmp/rev/Sample5_noninterleaved.1.fastq, ' +
+                               'filename Sample5_noninterleaved.1.fastq')
```
MrCreosote: This error message doesn't make sense - it's saying shock nodes but giving file names. It needs to provide shock nodes to make debugging possible.

test/ReadsUtils_server_test.py
```diff
@@ -1050,22 +1100,41 @@ def test_upload_fail_interleaved_for_single(self):
...
+                                'rev_id': ret2['id'],
+                                'sequencing_tech': 'seqtech-pr1',
+                                'wsname': self.ws_info[1],
+                                'name': 'pairedreads1',
+                                'interleaved': 0},
+                               'Reading FASTQ record failed - non-blank lines are not a ' +
+                               'multiple of four. ' +
+                               'Shock node /kb/module/work/tmp/fwd/Sample5_noninterleaved.1.' +
+                               'missing_line.fastq, Shock filename ' +
+                               'Sample5_noninterleaved.1.missing_line.fastq')
```
MrCreosote, line 1236: Same as above - needs shock nodes in error message.

test/ReadsUtils_server_test.py
```diff
@@ -1845,15 +1914,6 @@ def test_bad_file_type_good_compress_ext(self):
...
 
-#     def test_no_file_info(self):
-#
-#         self.download_error(
-#             [self.getWsName() + '/no_file_info'],
-#             ('Unable to determine file type from Shock or Workspace ' +
-#              'data. Reads object no_file_info ({}). Shock node {}').format(
-#                 self.staged['no_file_info']['ref'],
-#                 self.staged['no_file_info']['fwd_node_id']))
-
```
MrCreosote, line 1856: Please don't delete this - at some point we should get this test to work.

MrCreosote (dismissed):
- There's still a few pep8 violations, looks like all of them are comments without whitespace between the # and the comment
- tests pass
- The spec documentation still needs to be updated to explain the new behavior
- The downloader still needs to be updated to return the new fields in the object

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: I find this style of doing things much more readable.

MrCreosote: My main issue was the separation of the interleaved logic into two parts lines apart when there's no need to do that and it makes understanding the interleaved logic harder. AFAIC you can use any if/else structure you want, just do the processing for the variable in one place when you can.

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: Pinged Mike and Bill about making it a task in the future to change EA utils sdk method's behavior.

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: Is it bad form to declare them above?

test/ReadsUtils_server_test.py
jkbaumohl: Will need todo later.

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: In that it's not necessary, has no effect, and is two effectively dead lines of code, yeah I guess

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: Done using the flake 8 ignore "# noqa"

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: Done using the flake 8 ignore "# noqa"

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: Done

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: Code is in Ranjan's private repo.  Mike agrees a task should be made, but this should not hold up the PR at this point.

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: Ok.  Changed in the spec.

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: OK, where's the task?

ReadsUtils.spec
```diff
@@ -193,6 +198,17 @@ module ReadsUtils {
             unavailable.
```
MrCreosote: We should add the clarification from the spec about read_size here.

ReadsUtils.spec
```diff
@@ -193,6 +198,17 @@ module ReadsUtils {
...
+        float read_length_mean - The mean read length. null if unavailable.
+        float read_length_stdev - The std dev of read length. null if unavailable.
+        string phred_type - Phred type: 33 or 64. null if unavailable.
+        int number_of_duplicates - Number of duplicate reads. null if unavailable.
+        float qual_min - Minimum Quality Score. null if unavailable.
+        float qual_max - Maximum Quality Score. null if unavailable.
+        float qual_mean - Mean Quality Score. null if unavailable.
+        float qual_stdev - Std dev of Quality Scores. null if unavailable.
+        mapping<string, float> base_percentages - percentage of total bases being 
+            a particular nucleotide.
```
MrCreosote: Is this null or empty if it's not available?

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -118,14 +119,14 @@ def _proc_upload_reads_params(self, params):
                              'with a local forward reads file')
         if not shock and revfile:
             revid = os.path.abspath(os.path.expanduser(revfile))
-        interleaved = 1 if params.get('interleaved') else 0
+        interleaved = 0
+        if params.get('interleaved') or revid:
+            interleaved = 1
         kbtype = 'KBaseFile.SingleEndLibrary'
         single_end = True
         if interleaved or revid:
```
MrCreosote: this if statement is logically identical to the one above, can combine and simplify

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,15 +862,51 @@ def upload_reads(self, ctx, params):
         self.log('Starting upload reads, parsing args')
         o, wsid, name, objid, kbtype, single_end, fwdid, revid, shock = (
             self._proc_upload_reads_params(params))
-        interleaved = 1 if (not single_end and not revid) else 0
+        #IF from shock fwdid and revid are shock nodes, if not shock they are file paths
+        dfu = DataFileUtil(self.callback_url)
+        fwdpath = None
+        revpath = None
```
MrCreosote, line 846: Same comment as the last two reviews: 
No need for these two lines, fwd and revpath are always set in the if/else below

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,15 +862,51 @@ def upload_reads(self, ctx, params):
...
+            fileinput = [{'shock_id': fwdid,
+                          'file_path': self.scratch + '/fwd/',
+                          'unpack': 'uncompress'}]
+            if revid:
+                fileinput.append({'shock_id': revid,
+                                  'file_path': self.scratch + '/rev/',
+                                  'unpack': 'uncompress'})
+            self.log('downloading reads file(s) from Shock')
+            files = dfu.shock_to_file_mass(fileinput)
+            shock = False
```
MrCreosote: Same comment as previous review:
No need for this line, shock is never used again

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,15 +862,51 @@ def upload_reads(self, ctx, params):
...
+            revname = None
+            fwdid = None
+            revid = None
+        if revpath:
+            #now interleave the files
+            intpath = os.path.join(self.scratch, self.get_file_prefix() + '.inter.fastq')
+            self.interleave(None, None, fwdname, fwdid,
+                            revname, revid, fwdpath, revpath, intpath)
+            fwdpath = intpath
+            revpath = None
```
MrCreosote: No need to set revpath to None here once next comment is done

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,15 +862,51 @@ def upload_reads(self, ctx, params):
...
+
+        interleaved = 1 if not single_end else 0
+
+        fhandle, rhandle, fsize, rsize = \
+            self.process_files_for_upload(fwdpath, revpath, interleaved)
```
MrCreosote: since there's no way the files aren't interleaved, there's no need to pass revpath into this method and the method itself can be greatly simplified.

test/ReadsUtils_server_test.py
```diff
@@ -1027,8 +1125,7 @@ def test_upload_fail_bad_fastq(self):
              'fwd_id': ret['id'],
              'name': 'bar'
              },
-            'Invalid fasta file /kb/module/work/tmp/fwd/Sample1_invalid' +
-            '.fastq from Shock node ' + ret['id'])
+            'Invalid fasta file /kb/module/work/tmp/fwd/Sample1_invalid.fastq')
```
MrCreosote: Same comment as previous review:
Not having the source shock node here makes errors impossible to debug

test/ReadsUtils_server_test.py
```diff
@@ -1050,22 +1147,44 @@ def test_upload_fail_interleaved_for_single(self):
              'fwd_id': ret['id'],
              'name': 'bar'
              },
-            'Invalid fasta file /kb/module/work/tmp/fwd/Sample5_interleaved' +
-            '.fastq from Shock node ' + ret['id'])
+            'Invalid fasta file /kb/module/work/tmp/fwd/Sample5_interleaved.fastq')
```
MrCreosote: Same comment as previous review:
Not having the source shock node here makes errors impossible to debug

test/ReadsUtils_server_test.py
```diff
@@ -1050,22 +1147,44 @@ def test_upload_fail_interleaved_for_single(self):
...
+                                'wsname': self.ws_info[1],
+                                'name': 'pairedreads1',
+                                'interleaved': 0},
+                               'Reading FASTQ record failed - non-blank lines are not a ' +
+                               'multiple of four. ' +
+                               'Shock node ' + ret1['id'] +
+                               ', Shock filename ' +
+                               'Sample5_noninterleaved.1.missing_line.fastq')
         self.delete_shock_node(ret1['id'])
         self.delete_shock_node(ret2['id'])
```
MrCreosote, line 1238: I realize we need duplicates of the tests above for reads files provided locally.

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -8,6 +8,7 @@
 from pprint import pformat
 from DataFileUtil.DataFileUtilClient import DataFileUtil
 from DataFileUtil.baseclient import ServerError as DFUError
+from kb_ea_utils.kb_ea_utilsClient import kb_ea_utils
```
MrCreosote, line 11: There's still 5-10 or so pep8 errors in the impl and test files, mostly due to no whitespace between # and the comment, although there are a couple of bad indents

MrCreosote (commented):
tests pass

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: Hopefully got all of these, flake8 does not pick these up

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: Not identical.  There is a rev_id check as well. I find the way I have the code more readable.

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: I prefer the declaration approach.  Conscience choice to keep.

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: Done

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: Need to set it here or the revpath will get set in self.process_files_for_upload(fwdpath, revpath, interleaved)
As a result lib2 will get populated.

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: The code works,  what if we decide to reverse the call on interleaving.  Wanted to keep the functionality.

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: Look again, they're logically identical. Having two logically equivalent if statements that are combinable in a row is not more readable than a single if statement.

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: I don't understand what you mean by declaration approach. In the current code the value fwdpath and revpath are always set later in the code, so initializing them here is pointless and confusing.

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: The point is that once the next comment is done process_... will not take the revpath parameter.

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: Increasing maintenance burden and code complexity for something that might happen in the future is not a reasonable approach. Always simplify if you can. Especially in this case since the now unnecessary more complex code is in the git history.

test/ReadsUtils_server_test.py
jkbaumohl: Done

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: Disagree.  Also rev_id not covered in the above.

jkbaumohl: done

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: done

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: done

test/ReadsUtils_server_test.py
jkbaumohl: There is no shock id since there is no longer "process_shock_ids_for_upload".
They are now local files because we did stats on them and potentially interleaved them.

test/ReadsUtils_server_test.py
jkbaumohl: There is no shock id since there is no longer "process_shock_ids_for_upload".
They are now local files because we did stats on them and potentially interleaved them.

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: J & I worked this out, turns out was a miscommunication wrt which if statement he thought I was talking about.

test/ReadsUtils_server_test.py
MrCreosote: The files originally came from shock, so there must be a shock ID.

test/ReadsUtils_server_test.py
MrCreosote: The files originally came from shock, so there must be a shock ID.

MrCreosote:
Tests fail: 

```
', u'KBH_61498', u'KBH_61499']
.........................................................................EE...
======================================================================
ERROR: test_upload_fail_paired_bad_fastq (ReadsUtils_server_test.ReadsUtilsTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "ReadsUtils_server_test.py", line 1171, in test_upload_fail_paired_bad_fastq
    ret2 = self.upload_file_to_shock('data/Sample_rev.fq')
  File "ReadsUtils_server_test.py", line 113, in upload_file_to_shock
    with open(os.path.abspath(file_path), 'rb') as dataFile:
IOError: [Errno 2] No such file or directory: '/kb/module/test/data/Sample_rev.fq'

======================================================================
ERROR: test_upload_fail_paired_bad_fastq_file (ReadsUtils_server_test.ReadsUtilsTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "ReadsUtils_server_test.py", line 1166, in test_upload_fail_paired_bad_fastq_file
    'REV Path : /kb/module/test/data/Sample_rev.fq.')
  File "ReadsUtils_server_test.py", line 954, in fail_upload_reads_regex
    self.impl.upload_reads(self.ctx, params)
  File "/kb/module/lib/ReadsUtils/ReadsUtilsImpl.py", line 882, in upload_reads
    revname, revid, fwdpath, revpath, actualpath)
  File "/kb/module/lib/ReadsUtils/ReadsUtilsImpl.py", line 371, in interleave
    with open(fwdpath, 'r') as f, open(revpath, 'r') as r:
IOError: [Errno 2] No such file or directory: '/kb/module/test/data/Sample_rev.fq'

----------------------------------------------------------------------
Ran 80 tests in 484.588s
```

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,15 +839,75 @@ def upload_reads(self, ctx, params):
         self.log('Starting upload reads, parsing args')
```
MrCreosote, line 840: ctx in the method parameters is unused, which is a pep8 violation

test/ReadsUtils_server_test.py
```diff
@@ -1,28 +1,29 @@
...
-try:
-    from ConfigParser import ConfigParser  # py2 @UnusedImport
-except:
-    from configparser import ConfigParser  # py3 @UnresolvedImport @Reimport
 
-from Workspace.WorkspaceClient import Workspace
-from Workspace.baseclient import ServerError as WorkspaceError
-from biokbase.AbstractHandle.Client import AbstractHandle as HandleService  # @UnresolvedImport @IgnorePep8
+import requests
+from biokbase.AbstractHandle.Client import AbstractHandle as HandleService
```
MrCreosote: This is an unresolved import, but doesn't actually cause anything to fail since the client is in the docker container

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,15 +839,75 @@ def upload_reads(self, ctx, params):
...
         o, wsid, name, objid, kbtype, single_end, fwdid, revid, shock = (
             self._proc_upload_reads_params(params))
-        interleaved = 1 if (not single_end and not revid) else 0
+        # IF from shock fwdid and revid are shock nodes, if not shock they are file paths
+        dfu = DataFileUtil(self.callback_url)
+        fwdpath = None
+        revpath = None
```
MrCreosote, line 846: Same comment as the last 3 reviews:
No need for these two lines, fwd and revpath are always set in the if/else below. These lines are redundant.

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote, line 840: upload_reads is now about 100 lines and > 2 pages of code, which is > 2x too long. Should separate into subroutines.

test/ReadsUtils_server_test.py
```diff
@@ -1027,8 +1137,9 @@ def test_upload_fail_bad_fastq(self):
              'fwd_id': ret['id'],
              'name': 'bar'
              },
-            'Invalid fasta file /kb/module/work/tmp/fwd/Sample1_invalid' +
-            '.fastq from Shock node ' + ret['id'])
+            'Invalid FASTQ file - Path: /kb/module/work/tmp/fwd/Sample1_invalid.fastq. ' +
+            'Input Shock ID : ',
+            do_startswith=True)
```
MrCreosote: Why do_startwith here? the shock id and file name are known.

test/ReadsUtils_server_test.py
```diff
@@ -1039,36 +1150,127 @@ def test_upload_fail_bad_fastq_file(self):
...
+        ret = self.upload_file_to_shock('data/Sample5_interleaved.fastq')
+        self.fail_upload_reads(
+            {'sequencing_tech': 'tech',
+             'wsname': self.ws_info[1],
+             'fwd_id': ret['id'],
+             'name': 'bar'
+             },
+            'Invalid FASTQ file - Path: /kb/module/work/tmp/fwd/Sample5_interleaved.fastq. ' +
+            'Input Shock ID : ',
+            do_startswith=True)
```
MrCreosote: Why do_startwith here? the shock id and file name are known.

test/ReadsUtils_server_test.py
```diff
@@ -1039,36 +1150,127 @@ def test_upload_fail_bad_fastq_file(self):
...
+        shutil.copy('data/' + revtf, revtarget)
+        self.fail_upload_reads({'fwd_file': fwdtarget,
+                                'rev_file': revtarget,
+                                'sequencing_tech': 'seqtech-pr1',
+                                'wsname': self.ws_info[1],
+                                'name': 'pairedreads1',
+                                'interleaved': 0},
+                               'Interleave failed - reads files do not have ' +
+                               'an equal number of records. ',
+                               do_startswith=True)
```
MrCreosote: Why do_startwith here? the incoming file names are known.

test/ReadsUtils_server_test.py
```diff
@@ -1039,36 +1150,127 @@ def test_upload_fail_bad_fastq_file(self):
...
+        shutil.copy('data/' + revtf, revtarget)
+        self.fail_upload_reads({'fwd_file': fwdtarget,
+                                'rev_file': revtarget,
+                                'sequencing_tech': 'seqtech-pr1',
+                                'wsname': self.ws_info[1],
+                                'name': 'pairedreads1',
+                                'interleaved': 0},
+                               'Reading FASTQ record failed - non-blank lines are not a ' +
+                               'multiple of four.',
+                               do_startswith=True)
```
MrCreosote: Why do_startwith here? the incoming file names are known.

MrCreosote (dismissed):
tests pass

MrCreosote:
passing this on to Mike; he can decide what to do with the remaining comments.

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: I kept it.  More of a style issue.  I like declared variables above.

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: Not addressed

jkbaumohl: This is addressed

lib/ReadsUtils/ReadsUtilsImpl.py
msneddon: https://kbase-jira.atlassian.net/browse/TASK-158

lib/ReadsUtils/ReadsUtilsImpl.py
msneddon: I don't think we should delete the 'ctx' variable (might have side effects?).  I'm not sure that we can actually get rid of this pep8 violation because 'ctx' is in the generated code, so we might just have to leave this as is.  Adding a comment to quiet the warning also isn't really possible because the warning probably appears in the generated code.

msneddon: I agree that the function should ideally be shortened, but I am ok with it as is for now.

In my opinion, the larger somewhat related refactor would be that the Impl file itself is too long, and most of the functionality should be pulled into separate files or classes.

MrCreosote: > I don't think we should delete the 'ctx' variable

Why not?

> (might have side effects?).

It can't possibly.

The only way I know of to silence this type of pep8 warning is to delete it or use it in some way, and deletion is the cleanest I've found. I do it in other modules without issues.

lib/ReadsUtils/ReadsUtilsImpl.py
msneddon: It is redundant, but it also doesn't bother me too much here one way or the other.  For readability, the better improvement I think would be to extract that following large IF clause into a separate function that gives returns the paths you need.  But I'm also ok with this as is.

lib/ReadsUtils/ReadsUtilsImpl.py
msneddon: I guess I need to improve my python chops :)  yes, looking into it, del looks like the canonical way to remove a variable from the function scope.  I at first thought it did something more... I could have lived with this pep8 warning, but I stand corrected- looks fine.

msneddon (approved):
There are places where refactoring would still improve readability, but overall looks pretty good to me, and good enough to accept.

PR #40, opened 2016-11-23
Changed to use new ea_utils method.

jkbaumohl:
Essentially moved ea_utils report parsing to ea_utils.  Now consume parsed report in dict form.  Also fixed a bug in the test.  Updated ea_utils client.


msneddon (approved):
Great!  looks good and tests pass for me.

PR #41, opened 2017-01-04
Added ability to give input reference for Reads generated from other Reads object inputs

jkbaumohl:
Sorry Gavin. guess I forgot to do this before I left.


ReadsUtils.spec
```diff
@@ -103,6 +103,15 @@ module ReadsUtils {
             single end reads.
         insert_size_std_dev - the standard deviation of the size of the
             genetic fragments. Ignored for single end reads.
+        sequencing_tech - the sequencing technology used to produce the
+            reads. (Is required if source_reads_ref is not specified)
```
MrCreosote: This should probably be higher up in the spec. I'd put it under the required section with a note like UNLESS source_reads_ref is specified.

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -126,32 +126,76 @@ def _proc_upload_reads_params(self, params):
...
+            # Check that it is a reads object. If not throw an eror.
+            single_input, kbasefile = self.check_reads(source_reads_object)
+            o = {'sequencing_tech': source_reads_object['data'].get("sequencing_tech"),
+                 'single_genome': source_reads_object['data'].get("single_genome")
+                 }
+            if 'strain' in source_reads_object['data']:
+                o['strain'] = source_reads_object['data']['strain']
+            if 'source' in source_reads_object['data']:
+                o['source'] = source_reads_object['data']['source']
+            if not single_input and not single_end:
```
MrCreosote: This doesn't make sense to me - isn't it an error to propagate paired end reads info to single end reads and vice versa? If xor(single_input, single_end) is true an error should be thrown, right?

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -126,32 +126,76 @@ def _proc_upload_reads_params(self, params):
...
+                                  "include").format(", ".join(parameters_should_unfilled)))
+            # Check that it is a reads object. If not throw an eror.
+            single_input, kbasefile = self.check_reads(source_reads_object)
+            o = {'sequencing_tech': source_reads_object['data'].get("sequencing_tech"),
+                 'single_genome': source_reads_object['data'].get("single_genome")
+                 }
+            if 'strain' in source_reads_object['data']:
+                o['strain'] = source_reads_object['data']['strain']
+            if 'source' in source_reads_object['data']:
+                o['source'] = source_reads_object['data']['source']
```
MrCreosote: Why don't these use the _add_field() function?

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -126,32 +126,76 @@ def _proc_upload_reads_params(self, params):
...
+            self._check_pos(ism, 'insert_size_mean')
+            issd = params.get('insert_size_std_dev')
+            self._check_pos(issd, 'insert_size_std_dev')
+            if not single_end:
+                read_orientation_out = 1 if params.get('read_orientation_outward') else 0
+                o.update({'insert_size_mean': ism,
+                          'insert_size_std_dev': issd,
+                          'interleaved': interleaved,
+                          'read_orientation_outward': read_orientation_out
+                          })
```
MrCreosote: This function is now way too long - needs to be split up between multiple sub functions.

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -126,32 +126,76 @@ def _proc_upload_reads_params(self, params):
...
+                # is a paired end input and trying to upload a filtered/trimmed
+                # paired end ReadsUtils. need to check for more fields.
+                ism = source_reads_object['data'].get('insert_size_mean')
+                issd = source_reads_object['data'].get('insert_size_std_dev')
+                read_orientation_out = source_reads_object['data'].get('read_orientation_outward')
+                o.update({'insert_size_mean': ism,
+                          'insert_size_std_dev': issd,
+                          'interleaved': interleaved,
+                          'read_orientation_outward': read_orientation_out
+                          })
```
MrCreosote: A lot of this code looks very similar to the code below - can it be split out into a common function?

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -126,32 +126,76 @@ def _proc_upload_reads_params(self, params):
...
+            parameters_should_unfilled = ['insert_size_mean', 'insert_size_std_dev',
+                                          'sequencing_tech', 'strain',
+                                          'source', 'read_orientation_outward']
+            if any(x in params for x in parameters_should_unfilled):
+                self.log(("'source_reads_ref' was passed, making the following list of " +
+                          "parameters {} erroneous to " +
+                          "include").format(",".join(parameters_should_unfilled)))
+                raise ValueError(("'source_reads_ref' was passed, making the following list of " +
+                                  "parameters : {} erroneous to " +
+                                  "include").format(", ".join(parameters_should_unfilled)))
```
MrCreosote: I'd do this check before fetching the object. No need to do the fetch if this check fails, and this check is much faster.

MrCreosote (changes requested):
Mostly code structure issues, although I think there is one bug. I haven't checked coverage/tests yet, I'll do that once these are addressed.

ReadsUtils.spec
jkbaumohl: Done. Note error if specified if source_reads_ref is specified

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: Done

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: No because when filtering paired end reads you can end up with 3 resulting reads.
A paired end read (where the pair representative passed in both direction), and a fwd and rev singleton end file (where one direction passed, but the other did not, thus resulting in a single end file result).

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: Done, please check that I am using it correctly.

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: Theoretically single to paired end should not happen that I know of, but wanted to cover it in case there is a situation I am missing.

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: Broke out into one additional function

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: I still think this is a good idea to do - the code setting up the new object is very similar in both cases.

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: Eh, the fn is only 3 lines, I'm sure you can figure it out and your unit tests should cover it anyway...

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: I still think this is a good idea to do - the code setting up the new object is very similar in both cases.

MrCreosote (commented):
There's still one comment unaddressed (added a new comment to the chain for that one) but I'm concerned about this:

> No because when filtering paired end reads you can end up with 3 resulting reads.
> A paired end read (where the pair representative passed in both direction), and a fwd and rev singleton end file (where one direction passed, but the other did not, thus resulting in a single end file result).

If I understand this correctly, when you filter a set of paired end reads, you could wind up with up to three new objects instead of just one - a PELibrary and 1-2 SELibraries, even though really it's the same library? This seems like a big issue - perhaps there should be a new type like FilteredPELib to handle this case, or new optional fields in PELibs to store filtered single end reads?
Otherwise, the single end portions of the filtered reads are going to get ignored by the rest of the workflow.

jkbaumohl:
Yes your interpretation is correct, up to 3 resulting objects is possible (1PE, and 2 SE). I spoke with Paramvir and this how he wants to handle it. You are correct it is possible that the SE read(s) will get ignored. Paramvir said that most of the time, people will not be interested in them (and probably are not applicable in a workflow).  I think the behavior is fine.  I do not think we need an new object type. From a ReadUtils upload perspective it just propagates the properties, the writer of the filtering methods will decide what objects get made. Per example in my PRINSEQ sdk method I make the PE reads with the name specified.  If there are SE reads made they get the same core name with something like "FWD_singletons", "REV_singletons" appended on them.

lib/ReadsUtils/ReadsUtilsImpl.py
jkbaumohl: Note these values are coming from different places.  The names of the variables are the same. But does not really merit a method for it.

MrCreosote: The point of a function is to do the same thing to values coming from different places. If you're doing the same thing in two places and the only difference is the incoming values, that's exactly where you want to abstract the behavior into a function or method

test/ReadsUtils_server_test.py
```diff
@@ -915,6 +915,211 @@ def test_interleaved_with_pe_inputs(self):
...
+             'interleaved': 1,
+             'wsname': self.ws_info[1],
+             'source_reads_ref': ref[0]['obj_ref'],
+             'name': 'foo'
+             },
+            ("The input reference reads is single end, that should not " +
+             "give rise to a paired end object."))
+        self.delete_shock_node(node)
+
     def check_lib(self, lib, size, filename, md5):
```
MrCreosote, line 1136: Needs a test to exercise the try / catch around fetching the source ref object with DFU, other than that looks good.

MrCreosote (dismissed):

MrCreosote:
Also noting here that with regard to the 1->3 objects issue, the decision was made to leave things as they are for now, but we'll want to fix it later.

MrCreosote:
Tests pass

PR #42, opened 2017-01-13
Add staging and web upload reads

Tianhao-Gu:
@msneddon @MrCreosote 

Could you please review? Thanks! 

PR Target: add web and staging area file upload_reads functionality.


lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -651,6 +827,7 @@ def __init__(self, config):
         self.scratch = config['scratch']
         self.callback_url = os.environ['SDK_CALLBACK_URL']
         self.ws_url = config['workspace-url']
+        self.token_user = os.environ['KB_AUTH_TOKEN'].split('client_id=')[1].split('|')[0]
```
Tianhao-Gu: I think it leaves a potential issue if we change token format in the future. But I not sure a right way to get current user_name.

MrCreosote: Get the token from ctx['token'] and the user id from ctx['user_id'].

Tianhao-Gu: @MrCreosote Thanks! using ctx['token']. By the way, in my local test, ctx['user_id'] is None. So I am still using parsing user_name from token.

MrCreosote: You can set up ctx any way you want in the tests: https://github.com/kbaseapps/ReadsUtils/blob/master/test/ReadsUtils_server_test.py#L47

The user id is set when running normally: https://github.com/kbaseapps/ReadsUtils/blob/master/lib/ReadsUtils/ReadsUtilsServer.py#L567

MrCreosote:
Recording here that Tian and I discussed merging the 3 upload methods in this PR into one upload method:
1) Keep fwd/rev_id and fwd/rev_file as is, but deprecate
2) Add fwd/rev_reads (or something) inputs
3) Add reads_source input that's in [shock, file, web, staging] (or something)
4) The method should pull reads from the appropriate place

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: here's an example: https://github.com/kbaseapps/DataFileUtil/blob/master/test/DataFileUtil_server_test.py#L32-L39

Tianhao-Gu:
@MrCreosote Could you please give it another review? Thanks!

ReadsUtils.spec
```diff
@@ -69,6 +69,13 @@ module ReadsUtils {
         - OR -
         fwd_file - a local path to the reads data file: either single end
             reads, forward/left reads, or interleaved reads.
+        - OR - 
+        fwd_file_url - a download link that contains reads data file:
+            either single end reads, forward/left reads, or interleaved reads.
+        download_type - download type ['Direct Download', 'FTP', 'DropBox', 'Google Drive']
+        - OR - 
+        fwd_staging_file_name - reads data file name in staging area:
+            either single end reads, forward/left reads, or interleaved reads.
```
MrCreosote, line 88: Clarify the text at the beginning of the documentation here to cover the new options - only covers the previous options currently.

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -82,9 +87,20 @@ def _check_pos(self, num, name):
...
                              'a forwards reads file must be specified')
+        if fwdurl:
+            download_type = params.get('download_type')
+            if not (fwdurl and download_type):
+                raise ValueError('Both download_type and fwd_file_url must be provided')
+            fwdfile = self._download_web_file(fwdurl, download_type)
+        
+        if fwdstaging:
+            fwdfile = self._download_staging_file(params.get('user_id'), fwdstaging)
+
```
MrCreosote: A possible issue with downloading the files here and having the rest of the code treat them as if they already existed locally is that the error messages you get back from problems later in the code might not be helpful - it'll just include the file name of the downloaded file, and not provide information regarding how to find the source of the file. If you look at the error handling in the main method below the errors are different based on whether the file source was in shock or from a local file; my feeling is that something similar needs to be done for the alternate file sources.

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -117,8 +135,16 @@ def _proc_upload_reads_params(self, params):
         if not shock and revid:
             raise ValueError('Cannot specify a reverse reads file in shock ' +
                              'with a local forward reads file')
```
MrCreosote: The code should check that <= 1 of the rev* sources are specified, and that if it is specified, it matches the fwd* source.

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -14,6 +14,11 @@
 from numbers import Number
 import six
 import uuid
+import urllib2
```
MrCreosote, line 17: I personally prefer requests but eh

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +668,223 @@ def get_fq_stats(self, reads_object, file_path):
...
 
+
+    def _get_staging_file_path(self, token_user, upload_file_name):
+        """
+        _get_staging_file_path: return staging area file path
+
+        directory pattern: /data/bulk/user_name/file_name
+
+        """
+        return '/data/bulk/%s/%s' % (token_user, upload_file_name)
```
MrCreosote: does the /data/bulk part need to be configurable?

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +668,223 @@ def get_fq_stats(self, reads_object, file_path):
...
+        
+        """
+        if download_type == 'Direct Download':
+            self._download_direct_download_link(file_url, copy_file_path)
+        elif download_type == 'DropBox':
+            self._download_dropbox_link(file_url, copy_file_path)
+        elif download_type == 'FTP':
+            self._download_ftp_link(file_url, copy_file_path)
+        elif download_type == 'Google Drive':
+            self._download_google_drive_link(file_url, copy_file_path)
```
MrCreosote, line 800: throw an 'unknown download type: {}' (or something) error here

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +668,223 @@ def get_fq_stats(self, reads_object, file_path):
...
+        if not os.path.exists(dstdir):
+            os.makedirs(dstdir)
+        copy_file_path = os.path.join(dstdir, tmp_file_name)
+
+        self._download_file(download_type, file_url, copy_file_path)
+
+        return copy_file_path
+
+
+    def _download_file(self, download_type, file_url, copy_file_path):
```
MrCreosote, line 783: This and its subfunctions seem generally useful enough that they should live in DataFileUtils

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +668,223 @@ def get_fq_stats(self, reads_object, file_path):
...
+
+        self._check_ftp_connection(self.ftp_user_name, self.ftp_password, self.ftp_domain, self.ftp_file_path, self.ftp_file_name)
+        
+        ftp_connection = ftplib.FTP(self.ftp_domain)
+        ftp_connection.login(self.ftp_user_name, self.ftp_password)
+        ftp_connection.cwd(self.ftp_file_path)
+
+        # .gz file handler 
+        # TODO: create separate zip file handler for all download types 
+        if self.ftp_file_name.endswith('.gz'):
```
MrCreosote, line 901: Why is gz only handled for this case?

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +668,223 @@ def get_fq_stats(self, reads_object, file_path):
...
+        copy_file_path: output file saving path
+
+        """
+        # translate dropbox URL for direct download
+        if "?" not in file_url:
+            force_download_link = file_url + '?raw=1'
+        else:
+            force_download_link = file_url.partition('?')[0] + '?raw=1'
+
+        try: online_file = urllib2.urlopen(force_download_link)
```
MrCreosote: This part of the fn is identical to the last part of the direct download fn, could make a subfn

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +668,223 @@ def get_fq_stats(self, reads_object, file_path):
...
+            self.ftp_user_name = 'anonymous'
+            self.ftp_password = 'anonymous@domain.com'
+            self.ftp_domain = re.search('ftp://(.+?)/', file_url).group(1)
+            self.ftp_file_path = file_url.partition('ftp://')[-1].partition('/')[-1].rpartition('/')[0]
+            self.ftp_file_name = re.search('ftp://.*/(.+$)', file_url).group(1)
+
+        self._check_ftp_connection(self.ftp_user_name, self.ftp_password, self.ftp_domain, self.ftp_file_path, self.ftp_file_name)
+        
+        ftp_connection = ftplib.FTP(self.ftp_domain)
+        ftp_connection.login(self.ftp_user_name, self.ftp_password)
```
MrCreosote, line 891: Er, do you really want to pass passwords over a cleartext ftp connection?

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +668,223 @@ def get_fq_stats(self, reads_object, file_path):
...
+        _get_staging_file_path: return staging area file path
+
+        directory pattern: /data/bulk/user_name/file_name
+
+        """
+        return '/data/bulk/%s/%s' % (token_user, upload_file_name)
+
+
+    def _download_staging_file(self, token_user, staging_file_name):
+        """
```
MrCreosote, line 738: It'd be good for this and the other download fns to log what they're doing

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +668,223 @@ def get_fq_stats(self, reads_object, file_path):
...
+                                    password: 'anonymous@domain.com'
+
+        params:
+        file_url: FTP download link
+        copy_file_path: output file saving path
+
+        """
+
+        # process ftp credentials 
+        ftp_url_format = re.match(r'ftp://.*:.*@.*/.*', file_url)
```
MrCreosote, line 865: Is there a lib that'll parse the url for you?

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +668,223 @@ def get_fq_stats(self, reads_object, file_path):
...
+            self.ftp_file_path = file_url.partition('ftp://')[-1].partition('/')[-1].rpartition('/')[0]
+            self.ftp_file_name = re.search('ftp://.*:.*@.*/(.+$)', file_url).group(1)
+        else:
+            self.ftp_user_name = 'anonymous'
+            self.ftp_password = 'anonymous@domain.com'
+            self.ftp_domain = re.search('ftp://(.+?)/', file_url).group(1)
+            self.ftp_file_path = file_url.partition('ftp://')[-1].partition('/')[-1].rpartition('/')[0]
+            self.ftp_file_name = re.search('ftp://.*/(.+$)', file_url).group(1)
+
+        self._check_ftp_connection(self.ftp_user_name, self.ftp_password, self.ftp_domain, self.ftp_file_path, self.ftp_file_name)
```
MrCreosote: Pretty sure this is too long for pep8 (we're using 100 chars as a limit)

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +668,223 @@ def get_fq_stats(self, reads_object, file_path):
...
+        file_url: Google Drive download link
+        copy_file_path: output file saving path
+
+        """
+        # translate Google Drive URL for direct download
+        force_download_link_prefix = 'https://drive.google.com/uc?export=download&id='
+        file_id = file_url.partition('/d/')[-1].partition('/')[0]
+        force_download_link = force_download_link_prefix + file_id
+
+        try: online_file = urllib2.urlopen(force_download_link)
```
MrCreosote: repeated code again

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: My feeling is that all the download code, whether from shock or elsewhere, should be consolidated in one part of the fn (using DFU as appropriate), and any errors thrown should identify the source of the data

MrCreosote (changes requested):
Haven't checked the tests yet, will do when current comments are addressed one way or another

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: Currently no. The mounted directory always starts with /data/bulk/

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: hmmm. urlparse doesn't do good job on ftp urls. Wasn't able to find any good source online either.

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: Embarrassed. :/ But Chris and Sunita requested a .gz support on this ftp downloader. I think most of the source has .gz file format. Currently they had to download to local and upload.

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: Fair enough

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: Probably want to separate gunzip from the d/l methods, and have it just run if necessary regardless of the source. I think only shock files are automatically uncompressed by DFU. There's a DFU.unpack_file() (or something like that) function you could use, although like I say it'd probably make sense to move this stuff into DFU eventually.

ReadsUtils.spec
Tianhao-Gu: Could you explain more on this? Sorry, I think I didn't get what you mean.

MrCreosote: See lines 56-60 - they assume only 2 input types are now available instead of 4 (plus 4 download types, so 7 really)

Tianhao-Gu: 👍

Tianhao-Gu:
@MrCreosote I think I have addressed all of your concerns. Expect moving download_file to DataFileUtils part. I think it makes sense to move it to DFU. But do it after contractor's meeting?

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: @Tianhao-Gu is this one addressed?

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: Addressing this after the demo is fine - can you just stick in something like #TODO Tian move to DFU

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: Is this one addressed?

Tianhao-Gu:
@MrCreosote consolidated all downloads into a subfn.

Tianhao-Gu:
@MrCreosote When you have a minute, could you give it another review? Thanks!

ReadsUtils.spec
```diff
@@ -55,10 +55,20 @@ module ReadsUtils {
         
         If local files are specified for upload, they must be uncompressed.
         Files will be gzipped prior to upload.
+
+        If web files are specified for upload, a download type one of
+        ['Direct Download', 'DropBox', 'FTP', 'Google Drive'] must be specified too. 
+        The downloadable file must be uncompressed (except for FTP, .gz file is acceptable).
```
MrCreosote, line 61: Why doesn't the decompression run regardless of source?

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -82,12 +91,27 @@ def _check_pos(self, num, name):
...
+            if not params.get('download_type'):
+                raise ValueError('Both download_type and fwd_file_url must be provided')
+            reads_source = 'web'
+            fwdid = fwdurl
+        elif fwdstaging:
+            reads_source = 'staging'
+            fwdid = fwdstaging
+        elif fwdid:
+            reads_source = 'shock'
+            shock = True
```
MrCreosote: reads_source and shock are redundant

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -108,17 +132,31 @@ def _proc_upload_reads_params(self, params):
...
-            raise ValueError('Specified both a local file and a shock node ' +
-                             'for the reverse reads file')
-        if shock and revfile:
-            raise ValueError('Cannot specify a local reverse reads file ' +
+        revurl = params.get('rev_file_url')
+        revstaging = params.get('rev_staging_file_name')
+        if sum(bool(e) for e in [revid, revfile, revurl, revstaging]) > 1:
+            raise ValueError('Cannot specify more than one rev file source')
+
+        if shock and any([revfile, revurl, revstaging]):
```
MrCreosote: Just handle shock the same way as the other types - I don't think there's a need to do anything special is there?

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -133,7 +171,7 @@ def _proc_upload_reads_params(self, params):
                                                      interleaved, single_end)
         else:
             o = self._build_up_reads_data(params, single_end)
-        return o, wsid, name, objid, kbtype, single_end, fwdid, revid, shock
+        return o, wsid, name, objid, kbtype, single_end, fwdid, revid, reads_source
```
MrCreosote: This fn is now 82 lines, should probably split into smaller fns for readability

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +680,330 @@ def get_fq_stats(self, reads_object, file_path):
...
+
+        staging_file_path = self._get_staging_file_path(token_user, staging_file_name)
+
+        self.log('Start downloading staging file: %s' % staging_file_path)
+        dstdir = os.path.join(self.scratch, 'tmp')
+        if not os.path.exists(dstdir):
+            os.makedirs(dstdir)
+        shutil.copy2(staging_file_path, dstdir)
+        copy_file_path = os.path.join(dstdir, staging_file_path)
+        self.log('Copied staging file from %s to %s' % (staging_file_path, copy_file_path))
```
MrCreosote: Just FYI the new improved way is 'Copied {} to {}'.format(v1, v2) but again, meh

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +680,330 @@ def get_fq_stats(self, reads_object, file_path):
...
+            raise ValueError('Invalid download type: %s. Please use one of %s' % (download_type, valid_download_type))
+
+        if download_type == 'Direct Download':
+            self._download_direct_download_link(file_url, copy_file_path)
+        elif download_type == 'DropBox':
+            self._download_dropbox_link(file_url, copy_file_path)
+        elif download_type == 'FTP':
+            self._download_ftp_link(file_url, copy_file_path)
+        elif download_type == 'Google Drive':
+            self._download_google_drive_link(file_url, copy_file_path)
```
MrCreosote, line 800: Still need to throw an error here

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +680,330 @@ def get_fq_stats(self, reads_object, file_path):
...
+
+        params:
+        file_url: FTP download link
+        copy_file_path: output file saving path
+
+        """
+        print "=== WARNING ===\n=== WARNING ===\n"
+        print "FTP IS NOT A SECURIED PROTOCOL\nCREDENTIALS AND FILE CONTENTS WILL BE EXPOSED TO THE PUBLIC"
+        print "\n=== STOP THE PROCESS NOW IF YOU DO NOT WISH TO EXPOSE PRIVATE DATA ==="
+        time.sleep(60)
```
MrCreosote: Sorry, I guess we miscommunicated here. I was thinking of just adding a note in the documentation / UI that ftp credentials would be exposed. Since we're not allowing creds any more I don't that's necessary, and definitely not a 60s pause for every job.

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +680,330 @@ def get_fq_stats(self, reads_object, file_path):
...
+            if self.ftp_user_name.lower() != 'anonymous':
+                raise ValueError("Currently we only support anonymous FTP")
+            self.ftp_password = file_url.rpartition('@')[0].rpartition(':')[-1]
+            self.ftp_domain = re.search('ftp://.*:.*@(.+?)/', file_url).group(1)
+            self.ftp_file_path = file_url.partition('ftp://')[-1].partition('/')[-1].rpartition('/')[0]
+            self.ftp_file_name = re.search('ftp://.*:.*@.*/(.+$)', file_url).group(1)
+        else:
+            self.log('Setting anonymous FTP user_name and password')
+            self.ftp_user_name = 'anonymous'
+            self.ftp_password = 'anonymous@domain.com'
```
MrCreosote, line 881: Is this supposed to be the literal domain.com or the actual domain parsed out in the next step?

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +680,330 @@ def get_fq_stats(self, reads_object, file_path):
...
+        ftp_connection.login(self.ftp_user_name, self.ftp_password)
+        ftp_connection.cwd(self.ftp_file_path)
+
+        ftp_copy_file_path = copy_file_path + '.gz' if self.ftp_file_name.endswith('.gz') else copy_file_path
+        with open(ftp_copy_file_path, 'wb') as output:
+            ftp_connection.retrbinary('RETR %s' % self.ftp_file_name, output.write)
+        self.log('Copied FTP file to: %s' % ftp_copy_file_path)
+
+        if self.ftp_file_name.endswith('.gz'):
+            self._unpack_gz_file(copy_file_path)
```
MrCreosote, line 902: Again, I'm not sure why we don't just move the decompression outside this method and decompress any input that needs it (other than Shock which is handled by DFU).

Also, DFU uses filemagic to detect gzipped files rather than extensions which is probably more robust, so when this is moved to DFU it should do the same thing.

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +680,330 @@ def get_fq_stats(self, reads_object, file_path):
...
+            revpath = self._download_web_file(rev, download_type) if rev else None
+        elif reads_source == 'staging':
+            # TODO: Tian move _download_staging_file to DFU
+            fwdpath = self._download_staging_file(user_id, fwd)
+            revpath = self._download_staging_file(user_id, rev) if rev else None
+        elif reads_source == 'local':
+            pass
+        else:
+            raise ValueError("Unexpected reads_source value. reads_source: %s" % reads_source)
+
```
MrCreosote, line 1013: So I'd just do the unzip here

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +680,330 @@ def get_fq_stats(self, reads_object, file_path):
...
+        returnVal = {'fwdpath': fwdpath, 
+                     'revpath': revpath,
+                     'fwdname': fwdname, 
+                     'revname': revname}
+
+        return returnVal
+
+    def _generate_validation_error_message(self, reads_source, actualpath, fwdpath, revpath, fwdname, revname, fwdid, revid, fwdurl, revurl, fwdstaging, revstaging):
```
MrCreosote: pep8 length

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -14,6 +14,12 @@
 from numbers import Number
 import six
 import uuid
+import urllib2
+from contextlib import closing
+import ftplib
+import re
+import gzip
+import time
```
MrCreosote: double import

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: Actually, I just opened the repo up in Eclipse and the pep8 checker shows a lot of pep8 errors. Can you run it through a pep8 checker and fix the issues?

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +680,330 @@ def get_fq_stats(self, reads_object, file_path):
...
+                validation_error_message += (
+                    " Input Shock IDs - FWD Shock ID : " +
+                    fwdid + ", REV Shock ID : " + revid +
+                    ". FWD File Name : " + fwdname +
+                    ". REV File Name : " + revname +
+                    ". FWD Path : " + fwdpath +
+                    ". REV Path : " + revpath + ".")
+            else:
+                validation_error_message += (" Input Shock ID : " + 
+                                        fwdid + ". File Name : " + fwdname + ".")
```
MrCreosote: Missing the file path

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +680,330 @@ def get_fq_stats(self, reads_object, file_path):
...
+        elif reads_source == 'web':
```
MrCreosote, line 1042: web & staging should include the file path as well

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +680,330 @@ def get_fq_stats(self, reads_object, file_path):
...
+            # TODO: Tian move _download_web_file to DFU
+            fwdpath = self._download_web_file(fwd, download_type)
+            revpath = self._download_web_file(rev, download_type) if rev else None
+        elif reads_source == 'staging':
+            # TODO: Tian move _download_staging_file to DFU
+            fwdpath = self._download_staging_file(user_id, fwd)
+            revpath = self._download_staging_file(user_id, rev) if rev else None
+        elif reads_source == 'local':
+            pass
```
MrCreosote: set fwdpath and revpath to fwdid and revid here to maintain the same interface as the rest of the types (see below)

Also no need to declare *path above if you do this

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -892,22 +1284,13 @@ def upload_reads(self, ctx, params):
...
-                        ". REV File Name : " + revname +
-                        ". FWD Path : " + fwdpath +
-                        ". REV Path : " + revpath + ".")
-                else:
-                    validation_error_message += (" Input Shock ID : " + fwdid +
-                                                 ". File Name : " + fwdname + ".")
-            elif revpath:
-                validation_error_message += (" Input Files Paths - FWD Path : " +
-                                             fwdpath + ", REV Path : " + revpath + ".")
+            validation_error_message = self._generate_validation_error_message(
```
MrCreosote, line 1336: Why not just pass in ret (and include fwdid and revid in ret) here? Then the whole block starting with if reads_source == 'shock' above isn't needed, I think.

MrCreosote: So it'd be _gen_val_err(reads_source, actualpath, ret) I think

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -642,6 +680,330 @@ def get_fq_stats(self, reads_object, file_path):
...
+            else:
+                validation_error_message += (" Input URL : " + fwdurl + ".")
+        elif reads_source == 'staging':
+            if revstaging:
+                validation_error_message += (" Input Staging files - FWD Staging file : " +
+                                        fwdstaging + ", REV Staging file : " + revstaging + ".")
+            else:
+                validation_error_message += (" Input Staging : " + fwdstaging + ".")
+        elif reads_source == 'local':
+             if revpath:
```
MrCreosote: This this is missing an else for if there is not revpath

MrCreosote (changes requested):
So far I haven't been able to get tests to pass, but I suspect that's due to auth issues and not the tests themselves. Investigating. Submitting this review for now, will review test coverage later.

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

ReadsUtils.spec
Tianhao-Gu: I put a decompression for just FTP because it's just a temp solution for Chris and Sunita's test. Ideally, after contractor's demo, I can do a concrete decompression in FDU.

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: The domain in password can be anything. The password can be any email format. The actual domain is parsed in the next step.

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: yup yup. It's a very very temp thing for now. I am not happy this _unpack_gz_file either.

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: I did
"validation_error_message = "Invalid FASTQ file - Path: " + actualpath + ".""

at the beginning.

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: For local file path if not revpath, error message will just be

validation_error_message = "Invalid FASTQ file - Path: " + actualpath + "."

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

ReadsUtils.spec
MrCreosote: OK.

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: Oh right, you only need the fwd and rev paths for non-interleaved paired end reads.

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: Gotcha

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: hmmm I did this because I no longer want to treat fwdid as fwdurl/fwdstaging. I think it's more confusing to me.

MrCreosote: rename it to fwd_location or something - the point is that the combination of fwdid, revid, and reads_source tell you where to get a set of reads - think of it like an object where the api is fwdid and revid and it has different implementations based on the source, which is specified by reads_source. That approach makes the code way less complicated and results in many fewer variables.

Tianhao-Gu: 👍

Tianhao-Gu:
@MrCreosote Fixed PEP8 and other concerns. Could you please take another look when you have a sec? Thanks Thanks!

sychan:
If the module works reasonably, I would suggest we get this out and then batch up additional changes for a subsequent release. We need to get these uploaders into play ASAP.

Tianhao-Gu:
@MrCreosote Yea. It happens to me too. I just had to rerun tests couple of times till it all passes.

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -14,6 +14,11 @@
 from numbers import Number
 import six
 import uuid
+import urllib2
+from contextlib import closing
+import ftplib
+import re
+import gzip
```
MrCreosote, line 21: I still see ~10 pep8 errors, but meh for now

test/ReadsUtils_server_test.py
```diff
@@ -10,6 +10,8 @@
 from os import environ
 from pprint import pprint
 from zipfile import ZipFile
+import ftplib
+from mock import patch
```
MrCreosote, line 14: Lots of pep8 errors in this file

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -106,39 +111,92 @@ def _proc_upload_reads_params(self, params):
...
         else:
             o = self._build_up_reads_data(params, single_end)
-        return o, wsid, name, objid, kbtype, single_end, fwdid, revid, shock
+        return o, wsid, name, objid, kbtype, single_end, fwdsource, revsource, reads_source
+
+    def _check_rev_params(self, revid, revfile, revurl, revstaging, reads_source):
+        if sum(bool(e) for e in [revid, revfile, revurl, revstaging]) > 1:
+            raise ValueError('Cannot specify more than one rev file source')
+
+        if reads_source == 'shock' and any([revfile, revurl, revstaging]):
```
MrCreosote: Again, I'm not sure why shock is treated differently here - seems like this block could just be deleted and the error made the same type as the others? 

Also it seems like the next block isn't tested.

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -637,11 +698,359 @@ def validateFASTA(self, ctx, params):
...
+        reads_source: one of 'shock', 'web' or 'staging'
+        download_type: one of ['Direct Download', 'FTP', 'DropBox', 'Google Drive']
+        user_id: current token user
+        """
+
+        # return values
+        fwdname = None
+        revname = None
+        fwdpath = None
+        revpath = None
```
MrCreosote, line 978: Last 2 are guaranteed to be set below, no need to declare here

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -637,11 +698,359 @@ def validateFASTA(self, ctx, params):
...
+
+        if reads_source == 'shock':
+            # Grab files from Shock
+            fwdid = fwd
+            revid = rev
```
MrCreosote: Changing the variable names doesn't seem necessary for the few uses below

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -848,66 +1278,45 @@ def upload_reads(self, ctx, params):
...
+            fwdid = fwdsource
+            revid = revsource
 
         actualpath = fwdpath
         if revpath:
             # now interleave the files
-            actualpath = os.path.join(self.scratch, self.get_file_prefix() + '.inter.fastq')
+            actualpath = os.path.join(
+                self.scratch, self.get_file_prefix() + '.inter.fastq')
             self.interleave(None, None, fwdname, fwdid,
```
MrCreosote, line 1324: Unfortunately I just noticed this but interleave has the same error message issues as validation - right now it only will report shock nodes but not the locations of other files

test/ReadsUtils_server_test.py
MrCreosote, line 14: Tests passed 1st time

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -637,11 +698,359 @@ def validateFASTA(self, ctx, params):
...
+
+        self.log('Connecting and downloading web source: %s' % file_url)
+        try:
+            online_file = urllib2.urlopen(file_url)
+        except urllib2.HTTPError as e:
+            raise ValueError(
+                "The server couldn\'t fulfill the request.\nError code: %s" % e.code)
+        except urllib2.URLError as e:
+            raise ValueError("Failed to reach a server\nReason: %s" % e.reason)
+        else:
```
MrCreosote, line 822: Should add the url to the error string

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -637,11 +698,359 @@ def validateFASTA(self, ctx, params):
...
+        _download_dropbox_link: dropbox download link handler
+                                file needs to be shared publicly 
+
+        params:
+        file_url: dropbox download link
+        copy_file_path: output file saving path
+
+        """
+        # translate dropbox URL for direct download
+        if "?" not in file_url:
```
MrCreosote, line 839: I don't see a test for this case

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -637,11 +698,359 @@ def validateFASTA(self, ctx, params):
...
+
+        params:
+        file_url: FTP download link
+        copy_file_path: output file saving path
+
+        """
+        self.log('Connecting FTP link: %s' % file_url)
+        ftp_url_format = re.match(r'ftp://.*:.*@.*/.*', file_url)
+        # process ftp credentials
+        if ftp_url_format:
```
MrCreosote, line 867: I don't see a happy path test for this case

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -637,11 +698,359 @@ def validateFASTA(self, ctx, params):
...
+            try:
+                ftp.login(user_name, password)
+            except ftplib.all_errors, error:
+                raise ValueError("Cannot login: %s" % error)
+            else:
+                ftp.cwd(file_path)
+                if file_name in ftp.nlst():
+                    pass
+                else:
+                    raise ValueError("File %s does NOT exist in FTP path: %s" % (
```
MrCreosote, line 937: I don't see a test for this case

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -637,11 +698,359 @@ def validateFASTA(self, ctx, params):
...
+                    fwdsource + ", REV Shock ID : " + revsource +
+                    ". FWD File Name : " + fwdname +
+                    ". REV File Name : " + revname +
+                    ". FWD Path : " + fwdpath +
+                    ". REV Path : " + revpath + ".")
+            else:
+                validation_error_message += (" Input Shock ID : " +
+                                             fwdsource + ". File Name : " + fwdname + ".")
+        elif reads_source == 'web':
+            if revsource:
```
MrCreosote, line 1043: I don't see a test for this path

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -637,11 +698,359 @@ def validateFASTA(self, ctx, params):
...
+                validation_error_message += (" Input URLs - FWD URL : " +
+                                             fwdsource + ", REV URL : " + revsource +
+                                             ". FWD Path : " + fwdpath +
+                                             ". REV Path : " + revpath + ".")
+            else:
+                validation_error_message += (" Input URL : " + fwdsource + ".")
+        elif reads_source == 'staging':
+            if revsource:
```
MrCreosote, line 1051: I don't see a test for this path

MrCreosote (changes requested):

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍  removed this block.

Next block actually is tested in test_upload_fail_spec_fwd_file_rev_id.

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍 
remove fwdpath. 
revpath is not set when read_source == 'shock' and not revid.

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: @MrCreosote since I am logging all of my func activities, can I not touch this function? I am pretty sure users can find the url/staging file on the log.

MrCreosote: making the users parse through logs to figure out why their upload failed seems really user unfriendly to me. I try to put all the info they need to figure out what went wrong in the error.

Tianhao-Gu:
@MrCreosote Fixed interleave error message.

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote, line 21: Still ~10 pep8 errors and a bunch in the tests files. Seems like Tian's and my pep8 checker have different rules

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -106,39 +111,89 @@ def _proc_upload_reads_params(self, params):
...
-        return o, wsid, name, objid, kbtype, single_end, fwdid, revid, shock
+        return o, wsid, name, objid, kbtype, single_end, fwdsource, revsource, reads_source
+
+    def _check_rev_params(self, revid, revfile, revurl, revstaging, reads_source):
+        if sum(bool(e) for e in [revid, revfile, revurl, revstaging]) > 1:
+            raise ValueError('Cannot specify more than one rev file source')
+
+        if revid and reads_source != 'shock':
+            raise ValueError('Cannot specify a reverse reads file in shock ' +
+                             'with a local forward reads file')
```
MrCreosote: this error message is inaccurate - it may not be a local reads file. Should say missing shock forward path like the others

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -435,7 +493,18 @@ def interleave(self, source_obj_ref, source_obj_name, fwd_shock_filename,
                                                            rev_shock_node, rev_shock_filename])
                         error_message += 'Forward Path {}, Reverse Path {}.'
                         error_message_bindings.extend([fwdpath, revpath])
-                        raise ValueError(error_message.format(*error_message_bindings))
+
```
MrCreosote: The _read_fq_record funtion called above has the same issue with the errors only mentioning shock and local files

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -435,7 +493,18 @@ def interleave(self, source_obj_ref, source_obj_name, fwd_shock_filename,
                                                            rev_shock_node, rev_shock_filename])
                         error_message += 'Forward Path {}, Reverse Path {}.'
                         error_message_bindings.extend([fwdpath, revpath])
-                        raise ValueError(error_message.format(*error_message_bindings))
+                        
+                        if reads_source == 'web':
+                            error_message += 'Forward File URL {}, Reverse File URL {}.'
+                            error_message_bindings.extend([fwdsource, revsource])
+
+                        if reads_source == 'staging':
```
MrCreosote, line 514: I don't see a test for this path

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote, line 17: Oh, forgot - can you bump the version in kbase.yml

MrCreosote (changes requested):
Almost there. Forget about the pep8 changes if your checker says everything's fine, we'll figure out the differences later (if we get to it

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: yea. I am aware of this. But I need to mock _get_staging_file_path return. And for this function, _get_staging_file_path needs to return 2 different path. I cannot test it properly.

Tianhao-Gu:
@MrCreosote Addressed your concerns. Expect one missing test that I couldn't mock _get_staging_file_path to return different values.

lib/ReadsUtils/ReadsUtilsImpl.py
MrCreosote: Could you just set STAGING_FILE_PREFIX to something else prior to the test, or mock a fn that just returns the value of STAGING_FILE_PREFIX which is then used in _get_s_f?

test/ReadsUtils_server_test.py
```diff
@@ -2779,7 +2809,7 @@ def test_upload_reads_from_web_direct_download_paired_ends(self):
         self.assertEqual(d['read_orientation_outward'], 0)
         self.assertEqual(d['insert_size_mean'], None)
         self.assertEqual(d['insert_size_std_dev'], None)
-        self.check_lib(d['lib1'], 1341276, file_name, '04dac97530faa081c966e79af7c90883')
+        self.check_lib(d['lib1'], 1341276, file_name, '1c58d7d59c656db39cedcb431376514b')
```
MrCreosote: Why did the md5s change? Looking at the code changes I can't figure it out.

Tianhao-Gu: @MrCreosote It was a bug that I found out. 

previously when I called _download_web_file for rev file, I didn't pass rev_file=True argument.

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

Tianhao-Gu:
@MrCreosote YaY Fixed all

MrCreosote:
I think it would be better code if there were different class implementations of a common interface for the input types that abstracted away the input type behavior including generating error strings, but that can happen (much) later.

Tests pass.

PR #43, opened 2017-01-19
source_reads_ref should not cause error

dcchivian:
parameters_should_unfilled is too severe in its failure mode if there are duplicates.  If the value is the same as it is in the source, then don't fail!


MrCreosote:
Resolved via Slack

PR #44, opened 2017-01-23
Fix staging path

Tianhao-Gu:


Tianhao-Gu:
@MrCreosote Sorry. Had to release a hotfix. 

_download_staging_file returns staging area path instead of tmp directory path due to this mistake.

MrCreosote:
Why did tests not catch this, and do any tests need to be added/changed?

Tianhao-Gu:
@MrCreosote It breaks due to staging area is read only. In our tests, we mock _get_staging_file_path to return a local path. Therefore we don't have such issue.

Tianhao-Gu:
@MrCreosote Updated failing tests. Should have catch this before. My bad!

MrCreosote:
If you want to release on ci, the version will need a bump. If you only need dev/beta and don't worry about it

Tianhao-Gu:
@MrCreosote Thanks thanks. Bumped version. ^_^

MrCreosote:
Tests pass, but that *&)^_&*( insecure platform warning is back for me. Maybe it has something to do with rebuilding sdkbase...? not sure. This is probably the 3rd time I've fixed it

PR #45, opened 2017-01-24
add subdir support

Tianhao-Gu:


Tianhao-Gu:
@MrCreosote Could you please take a look at this PR? I made several improvements. 

1. Allowing users to have sub-directories in staging area. Previously I assume all files are in the root directory.
2. Give users a more meaningful error message if they input a improper download URL. 
3. Handle Google Drive's 2nd URL format. 

Thanks!

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -823,6 +825,14 @@ def _download_direct_download_link(self, file_url, copy_file_path):
             with closing(online_file):
                 with open(copy_file_path, 'wb') as output:
                     shutil.copyfileobj(online_file, output)
+            # check first 5 lines of file content
+            with open(copy_file_path) as copied_file:
+                for line in islice(copied_file, 5):
+                    if line.lower().find('html') != -1:
+                        raise ValueError("Undownable File.\n" + 
+                            "Please make sure file is publicly accessible\n" +
+                            "File URL: %s" % file_url)
```
MrCreosote: I don't understand this part at all - it looks like if the first 5 lines of the file contains `html` in any case the file is marked as invalid. That seems completely arbitrary - what if the reads file has `html` in the header somewhere? Also, once this gets moved to DFU, DFU could be downloading any file at all, in which case this check seems nonsensical.

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -725,24 +726,25 @@ def get_fq_stats(self, reads_object, file_path):
...
 
-    def _download_staging_file(self, token_user, staging_file_name):
+    def _download_staging_file(self, token_user, staging_file_subdir_path):
         """
         _download_staging_file: download staging file to scratch
 
         return: file path of downloaded staging file
 
         """
+        staging_file_name = staging_file_subdir_path.rpartition('/')[-1]
```
MrCreosote: why not just os.path.basename()?

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: @MrCreosote what about checking to see if first 5 lines has `<!DOCTYPE html>`. Or do you think we should just remove this check?

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: 👍

lib/ReadsUtils/ReadsUtilsImpl.py
Tianhao-Gu: @MrCreosote     <!DOCTYPE html>

MrCreosote: I guess first I need to understand what this code is for. Why do you want to exclude html files from download? It seems unlikely for our use case but it's possible someone might want to download an html file.

Tianhao-Gu: @MrCreosote I am trying to prevent user to input something like "www.google.com".Or in the case that the link is not publicly accessible, cloud API will return a webpage instead of the actual file. The code will continue executing till ValidateFASTQ find out it's not a fastq file format. 

In this downloader, downloader link should just be a fastq file. It shouldn't be a html file I think.

MrCreosote: OK, I see. Hmm. Even `html` isn't a valid test for all cases since it's optional. 

My gut feel is that we should just let the validator handle this case - is there good reason not to do that? Especially if we have an error message like `File X from url www.google.com is not a valid FASTQ file` the user should have a pretty clear idea of what went wrong.

Tianhao-Gu:
@MrCreosote Removed that html check. Rerunning all tests.

MrCreosote:
K, let me know when to re-review

Tianhao-Gu:
@MrCreosote All tests passed. ^_^

Tianhao-Gu:
@MrCreosote We comment on the same time. It should be good now. Thanks.

PR #48, opened 2017-01-28
Set seq_tech = Unknown when propagating from legacy type

MrCreosote:
...rather than throwing an error.

Also clean up pep8 stuff.


rsutormin (approved):
PR looks good

MrCreosote:
merging on roman's behalf

PR #49, opened 2017-01-31
Fixing missing rev_name in return dict of process_interleaved

samseaver:


msneddon:
This seems like it would cause tests to break- do any have to be updated?

samseaver:
I dont know, shouldn't tests have revealed this in the first place?

MrCreosote:
It looks to me like the current behavior is what's intended. All three of the process_* (paired, interleaved, single) methods do the same thing and return the original file names, not the new file names (which are available from the path). That is specified in the documentation: https://github.com/kbaseapps/ReadsUtils/blob/master/ReadsUtils.spec#L194

If we were to make this change, no information would be added to the returned data structure, and the original names of the files would be lost. Additionally, the change must be made in all three methods that produce this type of output.

The tests do indeed cover the file names, and now fail due to these changes.

samseaver:
OK, well, it threw an error because I was trying to call the filename of the de-interleaved files, and so, instead of being able to use

```rev_name=files['rev_name']```

I now use

```rev_name=files['rev'].split('/')[-1]```

But I thought this was a bug.

MrCreosote:
I'd suggest `os.path.basename()` but yeah, that's the way to do it.

PR #50, opened 2017-02-01
Fix handle bug

MrCreosote:
Handles might not contain the filename, but the code assumes they always
do.

Tests without impl fix reproduced the problem.


msneddon (approved):

msneddon:
Looks good.  We will need a version bump too to release this...

PR #53, opened 2017-02-24
using DFU

Tianhao-Gu:


Tianhao-Gu:
@MrCreosote Could you please take a look at this PR when you have a minute? Thanks. It should be fast, just remove all the download stuff.

MrCreosote:
Are staging files tested at all? 

It looks like all the web d/l types are tested, right?

Tests pass.

Tianhao-Gu:
@MrCreosote I need to mock a private func in DFU in order to test it. Since tests doesn't have access to staging area. But DFUClient doesn't have access to private func in DFU. Do you have a better idea to test it? It's all tested in DFU. Maybe we can skip the tests here?

All Web download types are tested here.

MrCreosote:
Hmm, I really don't like the idea of the staging logic path being untested - too easy for someone who doesn't completely understand the code to break it without realizing it.

Can you mock the return from the DFU client at the class level? e.g.

`mock(readsUtilsInstance.DataFileUtil.download_staging_file)`

 or does the DFU client instance itself have to be mocked?

Tianhao-Gu:
@MrCreosote Should have thought about this. I mocked DFUClient.download_staging_file func.

PR #56, opened 2017-07-20
Tests that check for compressed file size should check range

msneddon:
Different GZip compression algorithms create files with different sizes. The tests previously checked for an exact number of bytes, which break when the compression algorithm or library changes. That just happened so I switched the tests to check a small window instead of an exact size.

### DIFF
--- a/test/ReadsUtils_server_test.py
+++ b/test/ReadsUtils_server_test.py
@@ -1207,7 +1207,11 @@ class ReadsUtilsTest(unittest.TestCase):
         file_md5 = hashlib.md5(open(path, 'rb').read()).hexdigest()
         libfile = lib['file']
         self.assertEqual(file_md5, md5)
-        self.assertEqual(lib['size'], size)
+        size_buffer = 0.15  # allow +-15% size to permit a chagne in compression algorithm
+        min_size = int(float(size) * size_buffer - float(size))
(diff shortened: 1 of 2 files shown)


test/ReadsUtils_server_test.py
```diff
@@ -1207,7 +1207,11 @@ def check_lib(self, lib, size, filename, md5):
         file_md5 = hashlib.md5(open(path, 'rb').read()).hexdigest()
         libfile = lib['file']
         self.assertEqual(file_md5, md5)
-        self.assertEqual(lib['size'], size)
+        size_buffer = 0.15  # allow +-15% size to permit a chagne in compression algorithm
+        min_size = int(float(size) * size_buffer - float(size))
```
MrCreosote, line 1211: This is a negative number

MrCreosote (changes requested):

test/ReadsUtils_server_test.py
jayrbolton: If the thing is supposed to be +-15% then maybe instead:
```p
min_size = int(float(size) * 0.85)
max_size = int(float(size) * 1.15)
```

dcchivian:
This PR is stale.  Closing.

PR #58, opened 2018-02-01
adding commentary about what file types are acceptable to the upload menus.

sean-mccorkle:


ui/narrative/methods/import_pe_reads/display.yaml
```diff
@@ -1,6 +1,11 @@
 name     : PAIRED END READS
 subtitle : none
-tooltip  : Upload a PairedEndLibrary from FASTQ file(s) into your Narrative.
+tooltip  : |
+           Upload a PairedEndLibrary from FASTQ file(s) into your Narrative.
+
+           Valid file extensions for FASTQ: .fastq, .fnq, .fq.
+           Gzipped (.gz, .gzip) or bzipped files (.bz, .bz2, .bzip, .bzip2) are acceptable.
+           (<i>Note: .zip files are not currently supported</i>)
```
MrCreosote, line 8: It's more that file archives are not supported, while compressed files are.

sean-mccorkle: .zip can be either, I think, right?  Which is an issue.

sean-mccorkle: I should have added this is in response to https://kbase-jira.atlassian.net/browse/KBASE-5151 and https://kbase-jira.atlassian.net/browse/PUBLIC-304

MrCreosote: Well, zip is always an archive, but it could only contain one file. The problem is that it may contain several. Same problem for .tgz

MrCreosote: Which isn't to say this can't be fixed, it just requires smarter code.

sean-mccorkle: *Well, zip is always an archive, but it could only contain one file.*

My error - I thought that zip could make a single, compressed file analogous to .gz. I agree about .tgz

*Which isn't to say this can't be fixed, it just requires smarter code.*

I agree with that.  I also think we could scan the files to determine their format (ala what the unix file command does).   My thinking was that this warning message was a temporary caveat to users for the time being)

MrCreosote: > My error - I thought that zip could make a single, compressed file analogous to .gz.

Really? Example? I've never seen this. 

> My thinking was that this warning message was a temporary caveat to users for the time being)

Yeah, I'm just wondering if we should generalize to file archives rather than zip files only

sean-mccorkle: Was just checking (I don't use zip very much) - you're right - zip necessarily creates an archive, even for one file.  I was confusing myself thinking about unzipping a file with no parent directory.

It seems like handling archives would be a nice feature, allowing users to upload a bunch of files in one shot. But the handling code would have to be pretty savvy.  (ie - how deep - just a single folder/directory or more?  what if some files are okay but others aren't, that kind of thing)

MrCreosote: yeah, that's probably why we didn't do it in the first place. Would have taken a goodish chunk of time and we wanted to get this out (this was one of the very first modules we wrote that worked with the callback server). Also difficult to figure out which files are matched if there are multiple files and paired end reads.

JamesJeffryes (approved):
While the archive vs compressed file distinction might make this a touch more succinct, the existing message still seems like an improvement so I'm merging.

PR #59, opened 2018-09-05
PTV-1129 fixing upper case file extension

Tianhao-Gu:
Fixing RU validateFASTQ/FASTA doesn't validate file with upper case extension correctly.


Tianhao-Gu:
@JamesJeffryes  I cannot figure out why the changes in file_size in check_lib. Best guess is some package changed from DFU. But the MD5 are the same. I think it shouldn't be a big issue.

Dockerfile
```diff
@@ -1,13 +1,14 @@
...
 
-# RUN apt-get update
+RUN apt-get clean
+RUN apt-get update
 
-RUN sudo apt-get install python-dev libffi-dev libssl-dev
+RUN sudo apt-get install -y python-dev libffi-dev libssl-dev
 RUN pip install cffi --upgrade
 RUN pip install pyopenssl --upgrade
 RUN pip install ndg-httpsclient --upgrade
```
JamesJeffryes: I think that all these lines should not be needed with sdkbase2

JamesJeffryes (commented):
I think we can strip down the Dockerfile a bit but the code changes look good

Dockerfile
```diff
@@ -7,7 +7,6 @@ MAINTAINER KBase Developer
 
 # RUN apt-get update
 
-RUN sudo apt-get install python-dev libffi-dev libssl-dev
 RUN pip install cffi --upgrade
```
JamesJeffryes: Actually I think this whole block can go.

JamesJeffryes (approved):
I think you can drop the whole block of updates I indicated in sdkbase2 but feel free to merge whenever

JamesJeffryes:
💯I don't have merge powers though

PR #65, opened 2020-02-28
removing Windows CRLF char

Tianhao-Gu:
Windows newline char CRLF fails FASTQ validator. 

https://kbase-jira.atlassian.net/browse/PTV-1377


MrCreosote:
This seems like a weird thing to have to fix manually. I'd assume there's some way to get python to automatically convert windows line feeds to unix - have you looked into that?

MrCreosote:
Might be useful: https://stackoverflow.com/a/2717154/643675

MrCreosote:
maybe just ditch the check then fix optimization and just always run it through the fix loop with the python line converter

Tianhao-Gu:
@MrCreosote Yea. Python `open(file_path)` can handle it just fine. But it's the 3rd party validator that throws an error:
https://github.com/kbaseapps/ReadsUtils/blob/master/lib/ReadsUtils/ReadsUtilsImpl.py#L895-L897

MrCreosote:
that's why I'm saying to always run the fix loop, so you always remove blank lines and windows LF chars. Doing that is probably a whole lot faster than downloading the file so I doubt it'll add much time to the overall process

MrCreosote:
Should probably file a bug with the 3rd party dev too

Tianhao-Gu:
@MrCreosote sure. Just update code to always run fix loop.

kbase.yml
```diff
@@ -9,7 +9,7 @@ service-language:
     python
 
 module-version:
-    0.3.12
+    0.3.13
 
 owners:
     [gaprice, msneddon, jkbaumohl]
```
MrCreosote: Add yourself

MrCreosote (changes requested):
LGTM other than 1 thing

lib/ReadsUtils/ReadsUtilsImpl.py
```diff
@@ -862,28 +862,19 @@ def validateFASTQ(self, ctx, params):
...
-                        l = l.strip()
-                        if l:
-                            t.write(l + '\n')
-                            c += 1
-                    s.close()
-                    t.flush()
-                    shutil.copy2(t.name, file_path)
+            self.log('Removing blank lines and CRLF characters if any')
+            with open(file_path) as s, tempfile.NamedTemporaryFile(
+                    mode='w', dir=self.scratch) as t:
```
sean-mccorkle, line 869: I think this is a big improvement in the code

sean-mccorkle (commented):
This is not a show stopper, but if it were me, I would rename the function ValidateFASTQ to ValidateAndCleanFASTQ since its doing some cleanup on the file as well.

Also (again) not a show stopper:   I know FASTA files were not included in this, and I don't know of any occurrences of ^M's in FASTA files being reported, but is it worth considering checking those as well for this condition?  Or should that wait for an actual occurrence to be reported?

MrCreosote:
> This is not a show stopper, but if it were me, I would rename the function ValidateFASTQ to ValidateAndCleanFASTQ since its doing some cleanup on the file as well.

It's not really though - it's making a temporary file with an arbitrary name and making the changes there and discarding it when done. It doesn't fix the original file or provide the path to the new file AFIAK.

MrCreosote:
> This is not a show stopper, but if it were me, I would rename the function ValidateFASTQ to ValidateAndCleanFASTQ

That changes the API, not worth it IMO. Updating the docs would be good.

sean-mccorkle:
> That changes the API, not worth it IMO. Updating the docs would be good.

That's a good point.  I hadn't realized it was exposed in the spec.

Tianhao-Gu:
@MrCreosote @sean-mccorkle Thanks guys. Updated doc and added myself as module owner.

MrCreosote (approved):
Lgtm

PR #77, opened 2022-03-16
using personal google account instead of edu account

Tianhao-Gu:


Tianhao-Gu:
@MrCreosote @jkbaumohl getting lucky? tests all passed >_<

MrCreosote:
Do you know why the institutional account doesn't work?

Tianhao-Gu:
@MrCreosote hmmm I don't have a lot of clues.  There is an extra security setting saying `The organization's settings prevent changing this setting`. I am not sure what it means. But that's something I noticed is different than my personal google account.

MrCreosote (approved):
LGTM then

PR #80, opened 2024-03-11
Fix build, ignore google download tests

MrCreosote:
Google download doesn't seem to work currently.

Also
* Updated service auth client. There's something strange here - the authclient didn't get updated on a compile, so I did it manually.
* Deleted the unused baseclient in lib
* Just fetched the one jar from `jars` rather than cloning the whole thing
* Got rid of `six`


MrCreosote:
Tests passed prior to force push, but coverage push failed because the Makefile needs to be updated, oops. Force pushed to get rid of GHA codecov block

Tianhao-Gu (approved):
👍

PR #84, opened 2024-04-11
SCT-3242: add min_read_length

Tianhao-Gu:


test/ReadsUtils_server_test.py
```diff
@@ -680,6 +687,46 @@ def test_FASTQ_val_fail_bad_ext(self):
...
+        # The min_Sample.fastq file contains only one read with a length of 1.
+        read_length = 1
+
+        self.fail_upload_reads(
+            {'fwd_file': 'data/min_Sample.fastq',
+             'sequencing_tech': 'seqtech',
+             'wsname': self.ws_info[1],
+             'name': 'filereads1',
+             'min_read_length': read_length + 1},
+            "Invalid FASTQ file - Path: /kb/module/test/data/min_Sample.fastq."
```
MrCreosote, line 701: doesn't the exception give you any info about the problem?

test/ReadsUtils_server_test.py
```diff
@@ -680,6 +687,46 @@ def test_FASTQ_val_fail_bad_ext(self):
...
+                       'sequencing_tech': 'seqtech',
+                       'wsname': self.ws_info[1],
+                       'name': 'filereads1'})
+        obj = self.dfu.get_objects(
+            {'object_refs': [self.ws_info[1] + '/filereads1']})['data'][0]
+        self.assertEqual(ret[0]['obj_ref'], self.make_ref(obj['info']))
+        self.assertEqual(obj['info'][2].startswith(
+            'KBaseFile.SingleEndLibrary'), True)
+        d = obj['data']
+        self.assertEqual(d['read_count'], read_length)
```
MrCreosote: Why would the number of reads be equal to the min read length parameter?

test/ReadsUtils_server_test.py
```diff
@@ -680,6 +687,46 @@ def test_FASTQ_val_fail_bad_ext(self):
...
+        self.fail_upload_reads(
+            {'fwd_file': 'data/min_Sample.fastq',
+             'sequencing_tech': 'seqtech',
+             'wsname': self.ws_info[1],
+             'name': 'filereads1',
+             'min_read_length': read_length + 1},
+            "Invalid FASTQ file - Path: /kb/module/test/data/min_Sample.fastq."
+            )
+
+    def test_upload_min_len_reads(self):
```
MrCreosote, line 715: There should be at happy and sad tests that handle a case where the min length is something other than 1

and with > 1 read / file

test/ReadsUtils_server_test.py
MrCreosote, line 702: doesn't the exception give you any info about the problem?

test/ReadsUtils_server_test.py
Tianhao-Gu: 👍 you are right. They have no relationship.

test/ReadsUtils_server_test.py
Tianhao-Gu: no. validateFASTQ returns 0 and `_generate_validation_error_message` creates the error message.

test/ReadsUtils_server_test.py
Tianhao-Gu: I did a `test_upload_fail_min_len_reads` (sad) and `test_upload_min_len_reads` (happy) test. Read length shouldn't be matter here. What cases are not covered in your option? I actually feel this model is quite over tested.

MrCreosote: A case other than when min len reads is == 2 and reads / file >  1

test/ReadsUtils_server_test.py
MrCreosote: Oh, it's from *that* jar. Ok. finding something better has been a TODO for 8 years

Although giveni it's a jar maybe we could hook into the actual java code to improve things

MrCreosote: :+1:

test/ReadsUtils_server_test.py
MrCreosote: :+1:

test/ReadsUtils_server_test.py
Tianhao-Gu: I made the test file having 2 reads now.

MrCreosote: well... that's good but now I think you should add tests with reads with a length of 1 and do the same tests.

Tianhao-Gu: 👍

MrCreosote (approved):
lgtm

